PR #1, opened 2020-05-20
Add ESI node/network commands

tzumainn:
This initial code commit contains some basic structural files, as well as ESI commands regarding nodes and networks with accompanying tests.

### DIFF
--- a/esiclient/tests/v1/test_node_network.py
+++ b/esiclient/tests/v1/test_node_network.py
@@ -1,0 +1,413 @@
+#   Licensed under the Apache License, Version 2.0 (the "License"); you may
+#   not use this file except in compliance with the License. You may obtain
+#   a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#   WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#   License for the specific language governing permissions and limitations
+#   under the License.
+#
+
+from osc_lib import exceptions
+
+from esiclient.tests import base
+from esiclient.tests import utils
+from esiclient.v1 import node_network
+
+
+class TestList(base.TestCommand):
+
+    def setUp(self):
+        super(TestList, self).setUp()
+        self.cmd = node_network.List(self.app, None)
+
+        self.port1 = utils.create_mock_object({
+            "uuid": "port_uuid_1",
+            "node_uuid": "node_uuid_1",
+            "address": "aa:aa:aa:aa:aa:aa",
+            "internal_info": {'tenant_vif_port_id': 'neutron_port_uuid_1'}
+        })
+        self.port2 = utils.create_mock_object({
+            "uuid": "port_uuid_2",
+            "node_uuid": "node_uuid_2",
+            "address": "bb:bb:bb:bb:bb:bb",
+            "internal_info": {}
+        })
+        self.port3 = utils.create_mock_object({
+            "uuid": "port_uuid_3",
+            "node_uuid": "node_uuid_2",
+            "address": "cc:cc:cc:cc:cc:cc",
+            "internal_info": {'tenant_vif_port_id': 'neutron_port_uuid_2'}
+        })
+        self.node1 = utils.create_mock_object({
+            "uuid": "node_uuid_1",
+            "name": "node1"
+        })
+        self.node2 = utils.create_mock_object({
+            "uuid": "node_uuid_2",
+            "name": "node2"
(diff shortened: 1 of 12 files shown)


esiclient/v1/node_network.py
```diff
@@ -0,0 +1,203 @@
...
+        if node.provision_state != ACTIVE:
+            raise exceptions.CommandError(
+                "Node {0} must be in the active state".format(node.name))
+
+        baremetal_ports = ironic_client.port.list(node=node_uuid, detail=True)
+        has_free_port = False
+        for bp in baremetal_ports:
+            if 'tenant_vif_port_id' not in bp.internal_info:
+                has_free_port = True
+                continue
```
naved001, line 143: if you have found one free port, why do you want to iterate over the rest of them? Wouldn't a `break` statement be more suitable in this case?

esiclient/v1/node_network.py
```diff
@@ -0,0 +1,203 @@
...
+                ironic_client.node.update(node_uuid, node_update)
+            ironic_client.node.set_provision_state(node_uuid, ADOPT)
+            if len(node_revert) > 0:
+                ironic_client.node.update(node_uuid, node_revert)
+            # reload node information
+            node = ironic_client.node.get(node_uuid)
+
+        if node.provision_state != ACTIVE:
+            raise exceptions.CommandError(
+                "Node {0} must be in the active state".format(node.name))
```
larsks, line 136: We should ensure that errors are called out as errors when displaying output to the operator.  Currently, this would display a message like:

```
Node foo must be in the active state
```

But it's not clear if this is a warning, an error, or just informational.  Better would be:

```
ERROR: Node foo must be in the active state
```

I'm not sure if the solution is updating the individual error messages, or to catch the `CommandError` somewhere else and modify the output there.

esiclient/v1/node_network.py
```diff
@@ -0,0 +1,203 @@
...
+                                    'value': "{\"boot_option\": \"local\"}",
+                                    'op': 'add'})
+                node_revert.append({'path': '/instance_info/capabilities',
+                                    'op': 'remove'})
+
+            if len(node_update) > 0:
+                ironic_client.node.update(node_uuid, node_update)
+            ironic_client.node.set_provision_state(node_uuid, ADOPT)
+            if len(node_revert) > 0:
+                ironic_client.node.update(node_uuid, node_revert)
```
larsks, line 130: Should the above be wrapped in a `try/finally` block to prevent changes from accidentally persisting if the `set_provision_state` call fails?  E.g.:

```
try:
            if len(node_update) > 0:
                ironic_client.node.update(node_uuid, node_update)
            ironic_client.node.set_provision_state(node_uuid, ADOPT)
finally:
            if len(node_revert) > 0:
                ironic_client.node.update(node_uuid, node_revert)
```

esiclient/v1/node_network.py
tzumainn: Good points! I've added "ERROR" to these messages; it looks like other clients simply throw CommandError as well.

esiclient/v1/node_network.py
tzumainn: Ah, makes sense! I'll update this.

tzumainn:
Thanks for the reviews! I've addressed the comments (I think). Let me know if I missed anything!

esiclient/tests/v1/test_node_network.py
```diff
@@ -0,0 +1,411 @@
...
+        self.app.client_manager.baremetal.port.list.\
+            return_value = [self.port1]
+
+        arglist = ['node1', 'test_network']
+        verifylist = []
+
+        parsed_args = self.check_parser(self.cmd, arglist, verifylist)
+
+        self.assertRaisesRegex(exceptions.CommandError,
+                               'Node node1 has no free ports',
```
naved001, line 351: quick question because I am not too familiar with `assertRaisesRegex`: now that you have changed the output message to start with `ERROR: `, does this still work?

esiclient/tests/v1/test_node_network.py
```diff
@@ -0,0 +1,411 @@
...
+            return_value = None
+
+        arglist = ['node1', 'test_network']
+        verifylist = []
+
+        parsed_args = self.check_parser(self.cmd, arglist, verifylist)
+
+        self.assertRaisesRegex(
+            exceptions.CommandError,
+            'Network test_network is not attached to node node1',
```
naved001, line 409: same thing here. (does this need to be updated?)

esiclient/tests/v1/test_node_network.py
tzumainn: Ah, it will still work (because of the the regexp), but there's no reason I can't be more specific. I'll update these checks - thanks!

PR #2, opened 2020-06-03
Allow attaching/detaching node to specified port

tzumainn:
The use of trunk ports forces us to include the option of attaching
a node to a specific port. Given that, it becomes cleaner to force
the user to specify the port to be detached.

### DIFF
--- a/esiclient/v1/node_network.py
+++ b/esiclient/v1/node_network.py
@@ -68,11 +68,11 @@ class List(command.Lister):
                     if len(neutron_port.fixed_ips) > 0:
                         fixed_ip = neutron_port.fixed_ips[0]['ip_address']
                     data.append([node.name, port.address,
-                                 network.name, fixed_ip])
+                                 neutron_port.name, network.name, fixed_ip])
             elif not parsed_args.network:
-                data.append([node.name, port.address, None, None])
+                data.append([node.name, port.address, None, None, None])
 
-        return ["Node", "MAC Address", "Network", "Fixed IP"], data
+        return ["Node", "MAC Address", "Port", "Network", "Fixed IP"], data
 
 
 class Attach(command.ShowOne):
@@ -87,9 +87,15 @@ class Attach(command.ShowOne):
             metavar="<node>",
             help=_("Name or UUID of the node"))
         parser.add_argument(
-            "network",
+            "--network",
             metavar="<network>",
             help=_("Name or UUID of the network"))
+        parser.add_argument(
+            '--port',
+            dest='port',
+            metavar='<port>',
+            help=_("Attach to this port (name or UUID).")
+        )
 
         return parser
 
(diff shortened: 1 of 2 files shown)


esiclient/v1/node_network.py
```diff
@@ -87,23 +87,39 @@ def get_parser(self, prog_name):
...
+                "ERROR: You must specify either network or port")
 
         ironic_client = self.app.client_manager.baremetal
         neutron_client = self.app.client_manager.network
 
+        if parsed_args.network:
+            network = neutron_client.find_network(parsed_args.network)
+            port = None
+        if parsed_args.port:
+            port = neutron_client.find_port(parsed_args.port)
```
naved001, line 120: Just wanted to point out that the second `if` could actually be an `else` since you are already checking that one of  `parsed_args.network` or `parsed_args.port` must exist earlier in the code.
```python
        if parsed_args.network:
            network = neutron_client.find_network(parsed_args.network)
            port = None
        else: # we already checked if parsed_args.network does not exist then parsed_args.port must exist.
            port = neutron_client.find_port(parsed_args.port) 
```
But in the grand scheme of things it probably doesn't matter all that much, so I'll leave it as an optional suggestion. It just saves us one comparison.

esiclient/v1/node_network.py
```diff
@@ -149,17 +165,22 @@ def take_action(self, parsed_args):
...
+                node.name, port.name))
+            ironic_client.node.vif_attach(node_uuid, port.id)
+            network = neutron_client.get_network(port.network_id)
+        else:
+            print("Attaching network {1} to node {0}".format(
+                node.name, network.name))
+            port = neutron_client.create_port(name=node.name,
+                                              network_id=network.id)
+            ironic_client.node.vif_attach(node_uuid, port.id)
+            port = neutron_client.get_port(port.id)
```
naved001, line 179: probably a dumb question but it's not clear to me why are we either attaching a port to a node or a network to a node.

esiclient/v1/node_network.py
tzumainn: Good point! I'll update.

esiclient/v1/node_network.py
tzumainn: Solely for ease of use - it's more intuitive (I think) to be able to attach a network to a node and have the underlying code create a port for you. However things like trunk ports require being able to attach a network to an existing port. I could see the network-to-node option being dropped though!

naved001: > it's more intuitive (I think) to be able to attach a network to a node and have the underlying code create a port for you.

That makes sense.

> However things like trunk ports require being able to attach a network to an existing port. 

I think at some point we should have add a `docs` directory in this repo to document how to use the CLI.

tzumainn: 100% agree! I haven't done it yet because it feels like the CLI could still be in considerable flux. But I'll add a trello task to make sure it gets done.

PR #3, opened 2020-06-08
Initial doc

cudkowicz:


tzumainn:
This is a good first start, and admirably thorough! However, we actually want it to be a little *less* thorough; it's intended to be a quick reference for users; for that reason, there's no need for a lot of explanation. Try modeling it after something like: https://github.com/CCI-MOC/esi/blob/master/docs/usage.md

PR #4, opened 2020-06-09
initial doc

cudkowicz:


doc/commands.md
```diff
@@ -0,0 +1,33 @@
+# CLI commands documentation
+
+This document will cover and document three commands, attach, detach, and list. All three commands have a a log variable as well as two functions, get_parser, and take_action. 
+
+For each function the get_parser method is the same:
+* The function can add a node, network, or port, depending on the class. 
+* The parser is then returned. 
+
+Links to the commands are here:
```
tzumainn, line 9: We can probably cut all the text above here, leaving just the title

doc/commands.md
```diff
@@ -0,0 +1,33 @@
...
+* [Attach](#attach)
+* [Detach](#detach)
+* [List](#list)
+
+
+## <a name="attach"></a>Attach
+
+The attach command will attack a network to a node. 
+
+*`esi_node_network_attach node --port --network`, a node is required while a port and a network are optional.
```
tzumainn, line 19: The command a user would actually use is 'openstack esi node network attach'. I might display it as follows:

* `openstack esi node network attach <node> --port <port> --network <network>`
      * node: name or UUID of the node
      * port: name or UUID of the port
 
etc

You can get some information about what the parameters are in the code; I think there are helpstrings in there. Also note that the user has to specify exactly one of port or network.

doc/commands.md
```diff
@@ -0,0 +1,41 @@
...
+
+
+### <a name="attach"></a>Attach
+
+The attach command will attack a network to a node. 
+
+* `openstack esi node network attach <node> --port <port> --network <network>` 
+    * Exactly one port or network is needed.
+    * port: name or UUID of the port.
+    * node: name or UUID of the node.
```
tzumainn, line 22: Can you move `node` above `port`? This way the order of parameters matches.

doc/commands.md
```diff
@@ -0,0 +1,41 @@
...
+    * network: name or UUID of the network.
+* Returns the node, port address, fixed IP, IP address and which network was attached to the node.
+
+### <a name="detach"></a>Detach
+
+The detach command will detach a network from a node. 
+
+* `openstack esi node network detach <node> <port>`
+    * port: name or UUID of the port.
+    * node: name or UUID of the node.
```
tzumainn, line 32: Same here.

doc/commands.md
```diff
@@ -0,0 +1,41 @@
...
+
+### <a name="list"></a>List
+
+The list command will list the networks that are attached to a node. 
+
+* `openstack esi node network list --node <node> --network <network>`
+    * port: name or UUID of the port.
+    * node: name or UUID of the node.
```
tzumainn, line 40: And here. Also, this command doesn't take `port`; it takes `network`. It probably makes sense to also add some text explaining that these parameters filter the output.

doc/commands.md
cudkowicz: will do

doc/commands.md
cudkowicz: sure!

doc/commands.md
cudkowicz: I said that the given info determines the output.

PR #7, opened 2020-07-01
Add CLI commands for managing trunk ports

tzumainn:
This adds list/create/delete CLI commands for managing trunk ports,
giving ESI a simplified interface.

### DIFF
--- a/esiclient/v1/trunk.py
+++ b/esiclient/v1/trunk.py
@@ -1,0 +1,142 @@
+#   Licensed under the Apache License, Version 2.0 (the "License"); you may
+#   not use this file except in compliance with the License. You may obtain
+#   a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#   WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#   License for the specific language governing permissions and limitations
+#   under the License.
+
+import logging
+
+from osc_lib.command import command
+from osc_lib.i18n import _
+
+from esiclient import utils
+
+
+class List(command.Lister):
+    """List existing trunk ports and subports"""
+
+    log = logging.getLogger(__name__ + ".List")
+
+    def get_parser(self, prog_name):
+        parser = super(List, self).get_parser(prog_name)
+        return parser
+
+    def take_action(self, parsed_args):
+        self.log.debug("take_action(%s)" % parsed_args)
+
+        neutron_client = self.app.client_manager.network
+        trunks = neutron_client.trunks()
+
+        data = []
+        for trunk in trunks:
+            trunk_port = neutron_client.get_port(trunk.port_id)
+            network_names, port_names, fixed_ips \
+                = utils.get_full_network_info_from_port(
+                    trunk_port, neutron_client)
+            data.append([trunk.name,
+                         "\n".join(port_names),
+                         "\n".join(network_names)])
+
+        return ["Trunk", "Port", "Network"], data
+
+
+class Create(command.ShowOne):
+    """Create trunk port with subports"""
+
+    log = logging.getLogger(__name__ + ".Create")
+
+    def get_parser(self, prog_name):
+        parser = super(Create, self).get_parser(prog_name)
+        parser.add_argument(
+            "name",
+            metavar="<name>",
+            help=_("Name of trunk"))
+        parser.add_argument(
+            "--native-network",
+            metavar="<native_network>",
+            help=_("Name or UUID of the native network"))
+        parser.add_argument(
+            '--tagged-networks',
+            default=[],
+            dest='tagged_networks',
+            action='append',
+            metavar='<tagged_networks',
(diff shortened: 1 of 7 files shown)


esiclient/v1/trunk.py
```diff
@@ -0,0 +1,147 @@
...
+                = utils.get_full_network_info_from_port(
+                    trunk_port, neutron_client)
+            data.append([trunk.name,
+                         "\n".join(port_names),
+                         "\n".join(network_names)])
+
+        return ["Trunk", "Port", "Network"], data
+
+
+class Create(command.ShowOne):
```
naved001, line 53: Just curious, does it trigger anything on the switch?

esiclient/v1/trunk.py
```diff
@@ -0,0 +1,147 @@
...
+        parser.add_argument(
+            "name",
+            metavar="<name>",
+            help=_("Name of trunk"))
+        parser.add_argument(
+            "--network",
+            metavar="<network>",
+            help=_("Name or UUID of the primary network"))
+        parser.add_argument(
+            '--secondary-networks',
```
naved001, line 69: is it safe to assume that secondary networks will be attached as tagged networks on the switchport, and the primary network will be the native network?

esiclient/v1/trunk.py
```diff
@@ -0,0 +1,147 @@
...
+import logging
+
+from osc_lib.command import command
+from osc_lib.i18n import _
+
+from esiclient import utils
+
+ACTIVE = 'active'
+ADOPT = 'adopt'
+MANAGEABLE = 'manageable'
```
naved001, line 22: I didn't see them being used in this file. Is there any reason why we have these declared here?

esiclient/v1/trunk.py
```diff
@@ -0,0 +1,147 @@
...
+        return parser
+
+    def take_action(self, parsed_args):
+        self.log.debug("take_action(%s)" % parsed_args)
+
+        neutron_client = self.app.client_manager.network
+        trunk = neutron_client.find_trunk(parsed_args.name)
+
+        port_ids_to_delete = []
+        port_ids_to_delete.append(trunk.port_id)
```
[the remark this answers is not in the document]
tzumainn: Ah, good idea.

esiclient/v1/trunk.py
```diff
@@ -0,0 +1,147 @@
...
+        for sub_port in trunk.sub_ports:
+            port_ids_to_delete.append(sub_port['port_id'])
```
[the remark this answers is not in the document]
tzumainn: And this as well! I'll update.

esiclient/v1/trunk.py
tzumainn: Yep - trunk creation will trigger a switch update.

esiclient/v1/trunk.py
tzumainn: Yep - I'll rename these parameters to be clearer.

esiclient/v1/trunk.py
tzumainn: Whoops - accidental copy/paste. I'll remove!

tzumainn:
Thanks for the review! I've updated based on your feedback - let me know what you think!

esiclient/v1/trunk.py
```diff
@@ -0,0 +1,142 @@
...
+            metavar="<name>",
+            help=_("Name of trunk"))
+
+        return parser
+
+    def take_action(self, parsed_args):
+        self.log.debug("take_action(%s)" % parsed_args)
+
+        neutron_client = self.app.client_manager.network
+        trunk = neutron_client.find_trunk(parsed_args.name)
```
DanNiESh, line 134: If the argument is a non-exist trunk name, should it return directly?

tzumainn: Ah, my thought was that it would just return whatever error Neutron would return if the trunk doesn't exist. Do you think I should be more explicit?

DanNiESh:
A quick question, is an update command necessary? For example, update the trunk with different sub ports.

tzumainn:
> A quick question, is an update command necessary? For example, update the trunk with different sub ports.

Eventually, yes. However there are some unresolved issues around updating right now.

esiclient/v1/trunk.py
DanNiESh: I might suggest that because I don't know what kind of error Neutron would raise. Also, in creation, is it better to check if the trunk (name) already exists before creating a trunk twice?

DanNiESh:
> > A quick question, is an update command necessary? For example, update the trunk with different sub ports.
> 
> Eventually, yes. However there are some unresolved issues around updating right now.

Got it!

esiclient/v1/trunk.py
tzumainn: I think in general, my philosophy here is that we're just calling OpenStack commands and allowing whatever behavior that those OpenStack commands allow; we're not trying to override their capabilities. This way, if OpenStack functionality changes, we change right along with it. Let me know if you disagree!

DanNiESh: It makes sense to me now. We're calling OpenStack commands rather than creating new commands. Thanks for the explanation!

PR #8, opened 2020-07-07
This patch enables travis ci and runs some code linting tools

naved001:
I expect this to fail, just want to test that travis ci will run on pull requests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,11 @@
+language: python
+python:
+  - "3.8"
+
+install:
+  - pip3 install pycodestyle
+  - pip3 install pylint
+
+script:
+  - sh -e ci/pylint_tests.sh
+  - sh -e ci/pycodestyle_tests.sh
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -0,0 +1,11 @@
+language: python
+python:
+  - "3.8"
```
naved001, line 3: @tzumainn what's the version of python we should be using? I set it to 3.8 since that's what I had installed on my local machine.

naved001:
@tzumainn this one is ready for review. I am using pycodestyle (aka pep8) and pylint. pylint also checks for docstrings everywhere, I have disabled that for now to keep the size of this PR small.

.travis.yml
tzumainn: I'm actually unsure myself, but your PR doesn't break my checks on 3.7, so I'm fine with it! It can always be adjusted later.

PR #10, opened 2020-07-08
run tests in travis-ci

naved001:


naved001:
@tzumainn the travis environment says

> $ tox -e py37
> 
> /home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/tox/config/__init__.py:616: UserWarning: conflicting basepython version (set 3.7, should be 3.6) for env 'py36';resolve conflict or set ignore_basepython_conflict
> 
>   proposed_version, implied_version, testenv_config.envname,

It works, but why the warning, python version in the environment is 3.7 and that's the version I am using with tox. do you know what's going on?

naved001:
probably because we have py36 listed in our `tox.ini`
```
naved@naved-MS-7984:~/python-esiclient$ cat tox.ini |grep envlist
envlist = pep8,py37,py36
```

tzumainn:
Ah - yeah, we can probably remove that later.

PR #11, opened 2020-08-13
Add trunk add/remove network commands

tzumainn:
Adds additional CLI commands for adding and removing networks
from a trunk port.

### DIFF
--- a/esiclient/tests/v1/test_trunk.py
+++ b/esiclient/tests/v1/test_trunk.py
@@ -13,6 +13,8 @@
 
 import mock
 
+from osc_lib import exceptions
+
 from esiclient.tests import base
 from esiclient.tests import utils
 from esiclient.v1 import trunk
@@ -30,13 +32,13 @@ class TestList(base.TestCommand):
             "name": "trunk1-network1-trunk-port",
             "mac_address": "aa:aa:aa:aa:aa:aa",
         })
-        self.subport1 = utils.create_mock_object({
+        self.subport2 = utils.create_mock_object({
             "id": "port_uuid_2",
             "network_id": "network_uuid_2",
             "name": "trunk1-network2-sub-port",
             "mac_address": "bb:bb:bb:bb:bb:bb",
         })
-        self.subport2 = utils.create_mock_object({
+        self.subport3 = utils.create_mock_object({
             "id": "port_uuid_3",
             "network_id": "network_uuid_3",
             "name": "trunk1-network3-sub-port",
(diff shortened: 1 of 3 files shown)


esiclient/v1/trunk.py
```diff
@@ -113,6 +114,125 @@ def take_action(self, parsed_args):
...
+        self.log.debug("take_action(%s)", parsed_args)
+
+        tagged_networks = parsed_args.tagged_networks
+
+        if len(tagged_networks) == 0:
+            raise exceptions.CommandError(
+                "ERROR: no networks specified")
+
+        neutron_client = self.app.client_manager.network
+        trunk = neutron_client.find_trunk(parsed_args.name)
```
larsks, line 149: What happens if this fails (e.g., operator provided an invalid trunk name on the command line)?

tzumainn: Hm, it returns a weird error message. I'll catch this case.

Note that adding a network to a trunk where that network is already attached will result in a neutron error that's passed down appropriately, so I don't handle those cases. I guess it's just find_* methods that can return None and mess things up.

esiclient/tests/v1/test_trunk.py
```diff
@@ -328,3 +330,265 @@ def test_take_action(self):
...
+            "id": "network_uuid_2",
+            "name": "network2",
+            "provider_segmentation_id": 222
+        })
+        self.network3 = utils.create_mock_object({
+            "id": "network_uuid_3",
+            "name": "network3",
+            "provider_segmentation_id": 333
+        })
+        self.subport1 = utils.create_mock_object({
```
DanNiESh, line 351: Maybe name the subports as subport2 and subport3 would align with the ports and networks id?

esiclient/tests/v1/test_trunk.py
```diff
@@ -328,3 +330,265 @@ def test_take_action(self):
...
```
DanNiESh, line 481: Same as above.

esiclient/tests/v1/test_trunk.py
tzumainn: Fair enough! Should be fixed now.

esiclient/tests/v1/test_trunk.py
tzumainn: Fixed as well.

esiclient/v1/trunk.py
tzumainn: Okay, this should be fixed now!

tzumainn:
Thanks for the reviews! I've fixed the issues. Let me know if there are any more!

esiclient/v1/trunk.py
```diff
@@ -113,6 +114,133 @@ def take_action(self, parsed_args):
...
+
+        if trunk is None:
+            raise exceptions.CommandError(
+                "ERROR: no trunk named {0}".format(parsed_args.name))
+
+        sub_ports = []
+        for tagged_network_name in tagged_networks:
+            tagged_network = neutron_client.find_network(
```
larsks, line 157: Maybe do something similar here?

tzumainn: Oh, good point. Fixed!

PR #13, opened 2020-09-10
Create network and subnet together; create node and port together

DanNiESh:
Add new command to 
1. create the  baremetal node and port together 
2. create the network and subnet together

Passed unit tests and pep8 tests, haven't tested on openstack environment


esiclient/v1/network_subnet.py
```diff
@@ -0,0 +1,137 @@
...
+
+from osc_lib.cli import parseractions
+from osc_lib.command import command
+from osc_lib.i18n import _
+
+
+def _get_attrs_network(parsed_args):
+    attrs = {}
+    if parsed_args.network_name and \
+            parsed_args.network_name is not None:
```
larsks, line 23: This check is redundant (if `passed_args.network_name` is `None`, the first test will fail). You could just write:

```
if parsed_args.network_name:
    attrs["name"] = parsed_args.network_name
```

You can make the same change to several other conditionals in this file.

esiclient/v1/node_port.py
```diff
@@ -0,0 +1,123 @@
...
+            k, v = s.split("=", 1)
+        except ValueError:
+            raise exceptions.CommandError(
+                "ERROR: attributes must be a list of KEY=VALUE")
+        try:
+            v = json.loads(v)
+        except ValueError:
+            pass
+        dict[k] = v
+    return dict
```
larsks, line 62: Avoid naming your variables after native python types like `dict` or `list`, because invariably that will (a) confuse somebody and (b) eventually lead to weird bugs when you attempt to e.g. use the `dict()` constructor and instead find you're referring to your local variable.

esiclient/v1/node_port.py
```diff
@@ -0,0 +1,123 @@
...
+            attrs[k] = v
+    local_link_connection_values = attrs.get("local_link_connection")
+    if local_link_connection_values:
+        local_link_connection_dict = param_array_to_dict(
+            local_link_connection_values)
+        attrs.update({"local_link_connection": local_link_connection_dict})
+    return attrs
+
+
+def param_array_to_dict(param_array):
```
larsks, line 48: If the values are literals, you could replace this function with:

```
def param_array_to_dict(param_array):
    return dict(x.split('=', 1) for x in param_array)
```

If sometimes values will be a JSON object, you could instead write:

```
def param_array_to_dict(param_array):
    def _maybe_json(v):
        try:
            return json.loads(v)
        except ValueError:
            return v

    return dict((y[0], _maybe_json(y[1])) for y in [
        x.split('=', 1) for x in param_array
    ])
```

esiclient/v1/node_port.py
```diff
@@ -0,0 +1,123 @@
...
+#   under the License.
+
+import json
+import logging
+from osc_lib.command import command
+from osc_lib import exceptions
+from osc_lib.i18n import _
+
+
+def _get_attrs_node(parsed_args):
```
larsks, line 20: Consider instead:

```
def _get_attrs_node(parsed_args):
    field_list = ["name", "driver", "driver_info", "resource_class"]

    attrs = {
        k: v
        for k, v in vars(parsed_args).items()
        if k in field_list and v is not None
    }

    if 'driver_info' in attrs:
        attrs['driver_info'] = param_array_to_dict(attrs['driver_info'])

    return attrs
```

And similarly for other `_get_attrs_*` functions in this change.

setup.cfg
```diff
@@ -39,3 +39,4 @@ openstack.esiclient.v1 =
     esi_trunk_list = esiclient.v1.trunk:List
     esi_trunk_add_network = esiclient.v1.trunk:AddNetwork
     esi_trunk_remove_network = esiclient.v1.trunk:RemoveNetwork
+
```
larsks, line 42: This looks like an erroneous change.

esiclient/v1/network_subnet.py
DanNiESh: Fixed them!

esiclient/v1/node_port.py
DanNiESh: Thanks! I've changed the code. And I add another try except block to check x.split('=', 1).

esiclient/v1/node_port.py
DanNiESh: Fixed it!

setup.cfg
DanNiESh: Sorry, I forgot to add the new entry points in it...

joachimweyl:
@DanNiESh is this still a desired PR or should it be closed?

DanNiESh:
> @DanNiESh is this still a desired PR or should it be closed?

Nope, we can close it now.

PR #17, opened 2021-09-24
Delete existing volume connectors to prevent a conflict

tzumainn:
Creating a volume connector also implicitly creates a one-time
username/password through cinder that can't be retrieved or reused;
thus, using the same volume connector twice results in authentication
errors. Deleting and recreating volume connectors solves this issue.
Using a unique IQN each time prevents potential collisions as well.


DanNiESh (approved):
LGTM

PR #18, opened 2021-10-04
Funcitonal Tests

ljmcgann:
Adds base for functional tests. Adds first
node power tests


tzumainn (commented):
Looks good at first glance! A few quick comments:

* I think you're missing pytest in test-requirements
* I'd add some documentation in the README explaining how to run the tests, and what's required - specifically 'fake-hardware'. If there's existing documentation regarding how to activate the fake hardware in ironic, you can link to that; otherwise, I just looked at /etc/ironic/ironic.conf in your environment, noted every place were 'fake' was mentioned, and updated my own to match

.stestr.conf
```diff
@@ -1,3 +1,3 @@
 [DEFAULT]
-test_path=${TEST_PATH:-./esiclient/tests}
```
samuel-zuk, line 2: Since we use pytest in https://github.com/CCI-MOC/python-esileapclient, we should probably use pytest in this repo too (at least for the functional tests). I know the method I used to mark tests as slow assumed pytest as a test runner (although it may work fine under stestr too) and it would probably be nice regardless for consistency's sake.

esiclient/tests/functional/base.py
```diff
@@ -0,0 +1,188 @@
...
+                                   % (username, project_name, role))
+            self._cls_cleanups.append([admin_client.openstack, 'role remove',
+                                       '', '--user %s --project %s %s' %
+                                       (username, project_name, role)])
+
+            self._init_client(self, client_name, name)
+
+    def _verify_roles(self):
+        # Checks to see the is_node_owner and is_node_lessee roles exist
+        for role in 'owner', 'lessee':
```
samuel-zuk, line 170: Should we add a check for roles named `esi_leap_owner` and `esi_leap_lessee`? I know that in the [esi-leap default policy](https://github.com/CCI-MOC/esi-leap/blob/master/esi_leap/common/policy.py) having these be the names for the is_node_owner and is_node_lessee roles is also a valid configuration.

esiclient/tests/functional/base.py
```diff
@@ -0,0 +1,188 @@
...
+            self.clients['admin'].openstack('role show', '', '%s' % role)
+
+
+class CLIClient(base.CLIClient):
+
+    def ironic(self, action, flags='', params='', fail_ok=False,
+               merge_stderr=False):
+        return self.openstack('ironic %s' % action, '',
```
samuel-zuk, line 178: Do CLI commands of the form `$ openstack ironic [...]` exist? I was under the impression they were all of the form `$ openstack baremetal [...]`. If not, this method should probably be changed to `CLIClient.baremetal(...)` and if i'm wrong about this, a `CLIClient.baremetal(...)` method should probably be added.

esiclient/tests/functional/utils.py
```diff
@@ -0,0 +1,68 @@
...
+import json
+
+from tempest.lib.common.utils import data_utils
+from tempest.lib import exceptions
+
+
+def node_create(client, driver='fake-hardware', name=None, params='',
+                fail_ok=False, merge_stderr=False):
+
+    if not name:
```
samuel-zuk, line 24: We should probably be using data_utils.rand_name() regardless of whether or not one is specified to avoid conflicts with existing nodes on hardware these tests are to be run on. If a name is not specified, the name that is used in the CLI call will be something that looks like "tempest-baremetal-f45ab309cbb1". However, if a name is specified (e.g. "testing") the name used in the CLI call will just be "testing", rather than say, "tempest-testing-10a3bce11489f3".

esiclient/tests/functional/utils.py
```diff
@@ -0,0 +1,68 @@
...
+
+    output = client.openstack('baremetal node create '
+                              '--driver {0} --name {1} {2} -f json'
+                              .format(driver, name, params),
+                              ' ', ' ', fail_ok, merge_stderr)
+
+    return json.loads(output)
+
+
+def node_delete(client, identifier, ignore_exceptions=False,
```
samuel-zuk, line 35: `ignore_exceptions` here appears to serve the same functionality as `fail_ok`. This is my bad though; I probably should have documented the behavior of `fail_ok` better. See: https://opendev.org/openstack/tempest/src/branch/master/tempest/lib/cli/base.py

esiclient/tests/functional/utils.py
```diff
@@ -0,0 +1,68 @@
...
+from tempest.lib import exceptions
+
+
+def node_create(client, driver='fake-hardware', name=None, params='',
+                fail_ok=False, merge_stderr=False):
+
+    if not name:
+        name = data_utils.rand_name('baremetal')
+
+    output = client.openstack('baremetal node create '
```
samuel-zuk, line 27: Just so things are a bit cleaner, a `baremetal(...)` method should probably be added to the CLIClient class in base.py. This would make this call and calls like it a bit cleaner and would be a bit more consistent with the way it's done in the python-esileapclient repo (`client.openstack('baremetal node create', ...)` becomes `client.baremetal('node create', ...)`, etc.). It would also simplify the passing of positional arguments since if you use the client.openstack method you basically have to pass everything in as parameters which is a bit annoying sometimes.

esiclient/tests/functional/v1/test_node.py
```diff
@@ -0,0 +1,111 @@
...
+        utils.node_set(self.clients['admin'],
+                       identifier=node['name'],
+                       param='owner',
+                       value=self.projects['owner_proj']['id'])
+
+        utils.node_power_on(self.clients['owner_proj-owner'], node['name'])
+        utils.node_power_off(self.clients['owner_proj-owner'], node['name'])
+
+        self.assertRaises(exceptions.CommandFailed,
+                          utils.node_power_on,
```
samuel-zuk, line 57: This should probably be split off into another test (a negative test).

esiclient/tests/functional/v1/test_node.py
```diff
@@ -0,0 +1,111 @@
...
+
+        utils.node_set(self.clients['admin'],
+                       identifier=node['name'],
+                       param='lessee',
+                       value=self.projects['lessee_proj']['id'])
+
+        utils.node_power_on(self.clients['lessee_proj-lessee'], node['name'])
+        utils.node_power_off(self.clients['lessee_proj-lessee'], node['name'])
+
+        self.assertRaises(exceptions.CommandFailed,
```
samuel-zuk, line 87: This also should probably be split into a separate negative test case.

samuel-zuk (changes requested):
Looks good! I have some nitpicky comments, a few of which probably warrant further discussion but this looks very solid.

tox.ini
```diff
@@ -28,6 +28,10 @@ commands = flake8
 commands = {posargs}
 passenv = *
```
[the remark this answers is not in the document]
ljmcgann: Stuff like the _init_config I kinda just took for granted as just working. Looking a bit at it now, Id like to ask why we look at the virtualenv name?

esiclient/tests/functional/base.py
ljmcgann: Well, as it stands, we are only testing ironic stuff here. The policy file I configured with my ironic instance is here: https://github.com/CCI-MOC/esi/blob/master/etc/ironic/policy.yaml.sample. This policy file actually doesnt do anything with roles, just checking that certain fields (owner and lessee field of a node) are equal to the requester's project id

esiclient/tests/functional/base.py
ljmcgann: This code is wrong, but I dont use it anywhere. But youre right. I ended up preferring to have a more explicit function for each call we make, as that was how Ironic did things in their code.

tox.ini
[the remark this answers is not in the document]
samuel-zuk: When we run `$ tox -e functional` it installs all the dependencies and such to `./.tox/functional/bin/...`. If tests were run with a differently named venv it could cause a problem if the base class is looking for `./.tox/functional/bin/...` and stuff is installed to say, `./.tox/some_new_env/bin/...`

esiclient/tests/functional/utils.py
ljmcgann: I agree. I know I accidentally named it so after peeking into the Ironic functional tests, but for consistency's sake Ill change it

esiclient/tests/functional/v1/test_node.py
ljmcgann: I had thought of doing it like that, but for speed's sake I had went against it. Having to create and delete fewer nodes would be nice. But splitting it to negative tests would line up with what we have in the spreadsheet more I suppose

esiclient/tests/functional/base.py
samuel-zuk: Ah, that makes sense! Do you plan on having initialization error out if said roles don't exist on the cloud being tested?

.stestr.conf
ljmcgann: I changed this because I was having issues with the functional tests trying to run the unit tests. What would be the workaround to this?

tzumainn:
One more doc request - can you add comments per test similar to these?

https://github.com/CCI-MOC/python-esileapclient/blob/master/esileapclient/tests/functional/v1/test_basic.py#L26-L31

They're pretty prevalent across OpenStack, and really help people understand what the test is testing.

esiclient/tests/functional/utils.py
ljmcgann: im going to push back and say for the sake of getting something merged now, this is ok. Poor Mainn has been working off this pr and I'd hate to pull the rug from underneath him. Perhaps this can be a fixed in a followup pr

esiclient/tests/functional/v1/test_node.py
ljmcgann: agreed

esiclient/tests/functional/utils.py
tzumainn: Ah, feel free to fix this now :) I was wondering what the purpose of the clients in CLIClient were! I think all you'd have to do is rename the ironic client in CLIClient to baremetal, and do a replace-all of "client.openstack('baremetal" to "client.baremetal('"

ljmcgann: I didnt realize you could do client.baremetal. But if I'm understanding correctly Sam is suggesting having a more general options parser like in https://github.com/CCI-MOC/python-esileapclient/blob/c18f51a49107f4f87af555fde4fdca614717ee8c/esileapclient/tests/functional/utils/output_utils.py#L1

ljmcgann: I can refactor it this way, but going through specific baremetal and the eventual neutron CLIClient objects seemed overly complex to just spelling out exactly what we need in one function

esiclient/tests/functional/base.py
ljmcgann: dang. I guess I should gut this pr a bit then huh? Most of the work was in getting the dummy classes to work

tzumainn: I think the code is useful in general; at the very least users should probably have the 'member' role as that's baseline, and when we eventually move to secure RBAC we may require additional roles.

tzumainn (changes requested):
I think the one other thing I'd ask you to do is re-add the CLIClient class. It abstracts the various service CLIs, and I think it makes a lot of sense to call "self.baremetal" instead of "self.openstack". We don't need the flag parsing, but I like the service abstraction.

esiclient/tests/functional/base.py
```diff
@@ -0,0 +1,225 @@
...
+
+            admin_client.openstack('role add', '', '--user %s --project %s %s'
+                                   % (username, project_name, role))
+            self._cls_cleanups.append([admin_client.openstack, 'role remove',
+                                       '', '--user %s --project %s %s' %
+                                       (username, project_name, role)])
+
+            self._init_client(self, client_name, name)
+
+    def _init_random_client(self):
```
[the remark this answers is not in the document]
ljmcgann: It isnt quite as easy to do as you think, because of the weirdness with the class and instance method stuff in the base of the code, but I suppose it can be done.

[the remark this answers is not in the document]
tzumainn: What error do you get if you do what I suggest? Anyway, I'd recommend trying to get that to work because the amount of duplicated code is probably a bad sign.

esiclient/tests/functional/README.md
```diff
@@ -0,0 +1,16 @@
+# esi Client Functional Tests
+
+These tests are designed to ensure Ironic and Neutron behave as expected when configured with the Ironic policy file given here: [a link](https://github.com/CCI-MOC/esi/blob/master/etc/ironic/policy.yaml.sample).
+
+### Prerequisites
+
+These tests are intended to be ran against a functioning Openstack cloud with Ironic and Neutron services enables and running. It assumes that the ironic.conf file
+is located at `/etc/ironic/ironic.conf`. These tests assume that the `fake-hardware` hardware type is enabled under the `enabled_hardware_types` option under `[DEFAULT]` heading.
```
samuel-zuk, line 8: Two incredibly nitpicky things: Openstack should probably be formatted as OpenStack and also it might be useful to mention that you can pass in a custom Ironic cfg file via the `OS_IRONIC_CFG_PATH` env var for testing purposes.

esiclient/tests/functional/base.py
```diff
@@ -0,0 +1,173 @@
...
+            client_name = '%s-%s' % (name, role)
+            self.users[client_name] = {
+                'name': username,
+                'id': user_id,
+                'password': password,
+                'domain': 'default'
+            }
+
+            if role != 'member':
+                role = 'esi_leap_%s' % role
```
samuel-zuk, line 157: I think this is not something we want to add in the case of python-esiclient; i'd probably ditch this check entirely.

esiclient/tests/functional/utils.py
samuel-zuk: This still needs to be updated to prevent node name collisions, could probably look something like this:
```suggestion
    if not name:
        name = 'baremetal-node'
    name = data_utils.rand_name(name)
```

tox.ini
```diff
@@ -28,6 +28,11 @@ commands = flake8
...
 
+[testenv:functional]
+passenv = OS_*
+commands = pytest esiclient/tests/functional {posargs}
+VENV_NAME={envname}
```
samuel-zuk, line 34: I'm pretty sure this line needs to be:
```suggestion
setenv = VENV_NAME={envname}
```
for the env var to be set by tox

esiclient/tests/functional/utils.py
tzumainn: This issue only happens if the user is passing in the name, right? If so, I don't think we should be randomizing that name; what if the test is specifically trying to test the creation of a node with the same name?

esiclient/tests/functional/base.py
ljmcgann: I had last minute copy pasted part of the esi-leap code back in and missed this.

PR #22, opened 2021-11-04
Remove node-network-attach node state restriction

tzumainn:
Running the requisite node vif commands no longer depends on the
node state.


samuel-zuk (approved):
Looks good!

DanNiESh (commented):
Looks good! One tiny place: line 21-23 global variables could be removed in node_network.py.

tzumainn:
> Looks good! One tiny place: line 21-23 global variables could be removed in node_network.py.

Ooh, good catch! I've removed them.

DanNiESh (approved):
Thanks!

PR #23, opened 2021-11-04
Console functional tests

ljmcgann:
Adds tests to node to ensure protection of console
related apis.

### DIFF
--- a/esiclient/tests/functional/README.md
+++ b/esiclient/tests/functional/README.md
@@ -1,12 +1,12 @@
 # esi Client Functional Tests
 
-These tests are designed to ensure Ironic and Neutron behave as expected when configured with the Ironic policy file given here: [a link](https://github.com/CCI-MOC/esi/blob/master/etc/ironic/policy.yaml.sample). Alternatively, an Ironic config file path can passed via the evironment variable `OS_IRONIC_CFG_PATH`.
+These tests are designed to ensure Ironic and Neutron behave as expected when configured with the Ironic policy file given here: [https://github.com/CCI-MOC/esi/blob/master/etc/ironic/policy.yaml.sample]. Alternatively, an Ironic config file path can passed via the evironment variable `OS_IRONIC_CFG_PATH`.
 
 ### Prerequisites
 
 These tests are intended to be ran against a functioning OpenStack cloud with Ironic and Neutron services enabled and running. It assumes that the ironic.conf file
-is located at `/etc/ironic/ironic.conf`. These tests assume that the `fake-hardware` hardware type is enabled under the `enabled_hardware_types` option under `[DEFAULT]` heading.
-
+is located at `/etc/ironic/ironic.conf`. These tests assume that the `fake-hardware` hardware type is enabled under the `enabled_hardware_types` option under `[DEFAULT]` heading. Some of these tests require an Ironic node configured with ipmi credentials in order to test the use of the ipmitool-socat serial console.
+Details for configuring the serial console can be found here: [https://docs.openstack.org/ironic/latest/admin/console.html]
 Some tests require further configuration. This can be provided by doing the following:
 
 ```
--- a/esiclient/tests/functional/v1/test_node.py
+++ b/esiclient/tests/functional/v1/test_node.py
@@ -12,6 +12,8 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 
+import time
+
 import ddt
 from tempest.lib import exceptions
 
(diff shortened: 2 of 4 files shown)


esiclient/tests/functional/v1/test_node.py
```diff
@@ -137,3 +139,180 @@ def test_admin_can_node_power_on_off(self):
...
+        if node['provision_state'] != 'manageable':
+            utils.node_set_provision_state(self.clients['admin'],
+                                           self.node, 'manage')
+
+        utils.node_set(self.clients['admin'],
+                       self.node,
+                       'owner', self.projects['random']['id'])
+        self.addCleanup(utils.node_set,
+                        self.clients['admin'],
+                        self.node, 'owner')
```
tzumainn, line 228: You don't want to unset the owner; you want to set it to what it was previously by passing in node['owner']

esiclient/tests/functional/v1/test_node.py
```diff
@@ -137,3 +139,180 @@ def test_admin_can_node_power_on_off(self):
...
+        if node['provision_state'] != 'manageable':
+            utils.node_set_provision_state(self.clients['admin'],
+                                           self.node, 'manage')
+
+        utils.node_set(self.clients['admin'],
+                       self.node,
+                       'lessee', self.projects['random']['id'])
+        self.addCleanup(utils.node_set,
+                        self.clients['admin'],
+                        self.node, 'lessee')
```
tzumainn, line 274: Same here; we should set this to node['lessee']

esiclient/tests/functional/v1/test_node.py
```diff
@@ -137,3 +139,180 @@ def test_admin_can_node_power_on_off(self):
...
+        assert console['console_enabled']
+
+        utils.node_console_disable(self.clients['random-member'], self.node)
+        time.sleep(10)
+        console = utils.node_console_show(self.clients['random-member'],
+                                          self.node)
+        assert not console['console_enabled']
+
+    def test_random_cannot_enable_console(self):
+        """Tests lessee console functionality.
```
tzumainn, line 293: These comments may need to be updated

tzumainn (changes requested):
Looks pretty good! A few comments inline.

esiclient/tests/functional/v1/test_node.py
ljmcgann: fixed

esiclient/tests/functional/v1/test_node.py
ljmcgann: fixed

esiclient/tests/functional/v1/test_node.py
ljmcgann: fixed

ljmcgann:
Thinking about it, i think the sleeps might not actually be needed. We are only testing the protection of apis and not whether the console actually works. So it should be sufficient to just run the enable, disable, and show commands, and if they dont return an error, then we call it a success

esiclient/tests/functional/README.md
```diff
@@ -5,7 +5,7 @@ These tests are designed to ensure Ironic and Neutron behave as expected when co
 ### Prerequisites
 
 These tests are intended to be ran against a functioning OpenStack cloud with Ironic and Neutron services enabled and running. It assumes that the ironic.conf file
-is located at `/etc/ironic/ironic.conf`. These tests assume that the `fake-hardware` hardware type is enabled under the `enabled_hardware_types` option under `[DEFAULT]` heading.
+is located at `/etc/ironic/ironic.conf`. These tests assume that the `fake-hardware` hardware type is enabled under the `enabled_hardware_types` option under `[DEFAULT]` heading. Some of these tests require an Ironic node configured with ipmi credentials in order to test the use of the ipmitool-socat serial console.
```
tzumainn, line 8: May as well link to the Ironic documentation page describing what's needed.

esiclient/tests/functional/v1/test_node.py
tzumainn: Ah, I think you may have forgotten to update this test description!

tzumainn (commented):
Tests pass for me! I think the last two document-ish comments are the only other things that I'd like changed.

esiclient/tests/functional/README.md
ljmcgann: added

tzumainn (approved):
Looks good!

PR #24, opened 2021-11-22
Modify port creation to check if available port can be used

tzumainn:
This change modifies port creation to use an esi-specific name,
and avoids creating a port if a port with that name already exists
and is not in use.

### DIFF
--- a/esiclient/utils.py
+++ b/esiclient/utils.py
@@ -67,3 +67,26 @@ def get_full_network_info_from_port(port, client):
             port_names.append(subport.name)
 
     return network_names, port_names, fixed_ips
+
+
+def get_port_name(network, prefix=None, suffix=None):
+    port_name = network.name
+    if prefix:
+        port_name = "{0}-{1}".format(prefix, port_name)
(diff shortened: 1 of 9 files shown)


esiclient/utils.py
```diff
@@ -67,3 +67,26 @@ def get_full_network_info_from_port(port, client):
...
+        port_name = "{0}-{1}".format(prefix, port_name)
+    if suffix:
+        port_name = "{0}-{1}".format(port_name, suffix)
+    port_name = "esi-{0}".format(port_name)
+    return port_name
+
+
+def get_or_create_port(port_name, network, client):
+    ports = list(client.ports(name=port_name, status='DOWN'))
+    if len(ports) == 1:
```
DanNiESh, line 84: If there are multiple ports with the same name, why do we create a new port again?

tzumainn: That's a good point - I may as well just pick the first. Let me make that change!

DanNiESh (approved):
Looks good to me. Thanks!

PR #25, opened 2022-01-24
Adds Node List For Multisite

ljmcgann:
Adds first commands for multisite openstack
support

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,6 +31,7 @@ openstack.cli.extension =
     esiclient = esiclient.plugin
 
 openstack.esiclient.v1 =
+    esi_mdc_baremetal_node_list = esiclient.v1.mdc.mdc_node_baremetal:MDCBaremetalNodeList
     esi_node_network_attach = esiclient.v1.node_network:Attach
     esi_node_network_detach = esiclient.v1.node_network:Detach
     esi_node_network_list = esiclient.v1.node_network:List
--- a/esiclient/tests/unit/v1/mdc/test_mdc_node_baremetal.py
+++ b/esiclient/tests/unit/v1/mdc/test_mdc_node_baremetal.py
@@ -1,0 +1,129 @@
+#   Licensed under the Apache License, Version 2.0 (the "License"); you may
+#   not use this file except in compliance with the License. You may obtain
+#   a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#   WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#   License for the specific language governing permissions and limitations
+#   under the License.
+#
+
+import mock
+import munch
+
+from esiclient.tests.unit import base
+from esiclient.v1.mdc import mdc_node_baremetal
+
+
+class TestMDCBaremetalNodeList(base.TestCommand):
+
+    def setUp(self):
+        super(TestMDCBaremetalNodeList, self).setUp()
+        self.cmd = mdc_node_baremetal.MDCBaremetalNodeList(self.app, None)
+
+        class FakeConfig(object):
+
+            def __init__(self, name, region):
+                self._name = name
+                self.config = {'region_name': region}
+
+        self.cloud1 = FakeConfig('esi', 'regionOne')
+        self.cloud2 = FakeConfig('esi', 'regionTwo')
+        self.cloud3 = FakeConfig('esi2', 'regionOne')
+
+        self.node = munch.Munch(uuid="node_uuid_1", name="node1",
+                                instance_uuid="instance_uuid_1",
+                                power_state="off", provision_state="active",
+                                maintenance=False)
+
+        self.node2 = munch.Munch(uuid="node_uuid_2", name="node2",
+                                 instance_uuid="instance_uuid_2",
+                                 power_state="off", provision_state="active",
+                                 maintenance=False)
+
+        self.node3 = munch.Munch(uuid="node_uuid_3", name="node3",
+                                 instance_uuid="instance_uuid_3",
+                                 power_state="off", provision_state="active",
+                                 maintenance=False)
+
+        self.node4 = munch.Munch(uuid="node_uuid_4", name="node4",
+                                 instance_uuid="instance_uuid_4",
+                                 power_state="off", provision_state="active",
+                                 maintenance=False)
+
+    @mock.patch('esiclient.v1.mdc.mdc_node_baremetal.openstack.connect')
+    @mock.patch('esiclient.v1.mdc.mdc_node_baremetal.openstack.config.loader.'
+                'OpenStackConfig.get_all_clouds')
+    def test_take_action_list(self, mock_config, mock_connect):
+
+        mock_cloud = mock.Mock()
+        mock_cloud.list_machines.side_effect = [[self.node], [self.node2],
+                                                [self.node3, self.node4]]
+        mock_connect.return_value = mock_cloud
+
+        mock_config.return_value = [self.cloud1, self.cloud2, self.cloud3]
+
+        arglist = []
+        verifylist = []
+
+        parsed_args = self.check_parser(self.cmd, arglist, verifylist)
+
+        results = self.cmd.take_action(parsed_args)
+        expected = (
+            ['Cloud', 'Region', 'UUID', 'Name', 'Instance UUID', 'Power State',
+             'Provisioning State', 'Maintenance'],
+            [
+                [self.cloud1._name, self.cloud1.config['region_name'],
+                 "node_uuid_1", "node1", "instance_uuid_1", "off", "active",
+                 False],
+                [self.cloud2._name, self.cloud2.config['region_name'],
+                 "node_uuid_2", "node2", "instance_uuid_2", "off", "active",
+                 False],
+                [self.cloud3._name, self.cloud3.config['region_name'],
+                 "node_uuid_3", "node3", "instance_uuid_3", "off", "active",
+                 False],
+                [self.cloud3._name, self.cloud3.config['region_name'],
+                 "node_uuid_4", "node4", "instance_uuid_4", "off", "active",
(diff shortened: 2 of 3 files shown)


esiclient/v1/node_baremetal.py
```diff
@@ -0,0 +1,73 @@
...
+
+import logging
+
+import openstack
+from osc_lib.command import command
+from osc_lib import exceptions
+from osc_lib.i18n import _
+
+# Initialize and turn on debug logging
+openstack.enable_logging(debug=False)
```
tzumainn, line 21: this almost definitely should not be here. I think you may have turned on logging for the entire demo server by the way, as I see logs if I just log in and run 'openstack project list' - could you turn that off?

setup.cfg
```diff
@@ -35,6 +35,7 @@ openstack.esiclient.v1 =
     esi_node_network_detach = esiclient.v1.node_network:Detach
     esi_node_network_list = esiclient.v1.node_network:List
     esi_node_volume_attach = esiclient.v1.node_volume:Attach
+    esi_mdc_baremetal_node_list = esiclient.v1.node_baremetal:MDCBaremetalNodeList
```
tzumainn, line 38: These are generally in alphabetical order, so esi_mdc would go above all the esi_node entries.

esiclient/v1/node_baremetal.py
```diff
@@ -0,0 +1,73 @@
...
+                   "--clouds flag."
+                   )
+        )
+
+        return parser
+
+    def take_action(self, parsed_args):
+
+        columns = ['UUID', 'Name', 'Instance UUID', 'Power State',
+                   'Provisioning State', 'Maintenance']
```
tzumainn, line 57: We want the "data center" - essentially the cloud name used from clouds.yaml - as well

esiclient/v1/node_baremetal.py
```diff
@@ -0,0 +1,73 @@
...
+            '--clouds',
+            dest='clouds',
+            metavar='<clouds>',
+            nargs="+",
+            required=True,
+            help=_("Specify the cloud to use from clouds.yaml.")
+        )
+
+        parser.add_argument(
+            '--regions',
```
tzumainn, line 41: it looks like region is something that's part of clouds.yaml - can it be taken from there?

esiclient/v1/node_baremetal.py
```diff
@@ -0,0 +1,73 @@
...
+
+class MDCBaremetalNodeList(command.Lister):
+    """List baremetal nodes from multiple OpenStack instances"""
+
+    log = logging.getLogger(__name__ + ".List")
+
+    def get_parser(self, prog_name):
+        parser = super(MDCBaremetalNodeList, self).get_parser(prog_name)
+        parser.add_argument(
+            '--clouds',
```
tzumainn, line 32: I think this is a reasonable option, but can we also pull from an environment variable? This way the user only has to set it once.

esiclient/v1/node_baremetal.py
```diff
@@ -0,0 +1,73 @@
...
+
+        if len(parsed_args.clouds) != len(parsed_args.regions):
+            raise exceptions.CommandError(
+                "ERROR: Must specify one region per cloud.")
+
+        for i in range(len(parsed_args.clouds)):
+            cloud = openstack.connect(cloud=parsed_args.clouds[i],
+                                      region=parsed_args.regions[i])
+
+            nodes = cloud.list_machines()
```
tzumainn, line 68: This method is new to me, but it looks like a pretty good standardization. Interesting!

tzumainn (changes requested):
Good start! Don't forget unit tests :)

esiclient/v1/node_baremetal.py
tzumainn: I'd say that the user should have multiple clouds.yaml entries then, one for each region. Regardless if this is an option or not, we want the default to be as simple as possible for the end user, and having them list a region for each cloud they want to use isn't all that simple.

ljmcgann: It seems to be part of a normal workflow to have one cloud entry with multiple regions listed, the documentation for using the openstack.connect library gives an example using such a setup

tzumainn: In that case we should iterate over the regionName; if that's not possible then we should document that the CLI doesn't support multiple regions in a cloud. Having to list all the region names for multiple clouds each time a user runs the CLI isn't user-friendly, and that's the primary goal of this feature.

ljmcgann:
I have changed the command such that if no --clouds and --regions parameters are passed, all nodes from all clouds,region pairs in clouds.yaml will be displayed.

esiclient/v1/node_baremetal.py
```diff
@@ -0,0 +1,84 @@
...
+        data = []
+
+
+        if parsed_args.clouds is None and parsed_args.regions is None:
+            cloud_regions = openstack.config.loader.OpenStackConfig().get_all_clouds()
+            
+            for c in cloud_regions:
+
+                #cloud = openstack.connect(config=c)
+                cloud = openstack.connect(cloud=c._name, region=c.config['region_name'])
```
tzumainn, line 62: Is the name value retrievable from c.config? I'm slightly uncomfortable using an "_" attribute.

esiclient/v1/node_baremetal.py
```diff
@@ -0,0 +1,84 @@
...
+        parser.add_argument(
+            '--regions',
+            dest='regions',
+            metavar='<regions>',
+            nargs="+",
+            help=_("Specify the region to use from clouds.yaml. Must specify "
+                   "regions in order corresponding to clouds passed with "
+                   "--clouds flag."
+                   )
+        )
```
tzumainn, line 45: Let's just get rid of the region flag; I don't see why anyone would ever use it, and it cleans up the code below

esiclient/v1/node_baremetal.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def take_action(self, parsed_args):
+
+        columns = ['Cloud', 'Region', 'UUID', 'Name', 'Instance UUID', 'Power State',
+                   'Provisioning State', 'Maintenance']
+        data = []
+
+
+        if parsed_args.clouds is None and parsed_args.regions is None:
+            cloud_regions = openstack.config.loader.OpenStackConfig().get_all_clouds()
```
tzumainn, line 57: Assuming we get rid of the region parameter, can we change this to filter out the undesired clouds if parsed_args.clouds is passed in? That gets rid of the else statement, which currently leads to some logic duplication.

tzumainn (commented):
This works much better, thanks! Comments inline; reply on the comment if you fix them. Also, there's pep8 errors and the unit tests will need to be updated.

esiclient/v1/node_baremetal.py
ljmcgann: it is not

esiclient/v1/node_baremetal.py
ljmcgann: done

esiclient/v1/node_baremetal.py
ljmcgann: done

ljmcgann:
I removed the region flag and fixed the unit tests

esiclient/v1/node_baremetal.py
```diff
@@ -0,0 +1,61 @@
...
+        data = []
+
+        cloud_filter = parsed_args.clouds or []
+
+        cloud_regions = openstack.config.loader.OpenStackConfig().\
+            get_all_clouds()
+
+        for c in cloud_regions:
+
+            if cloud_filter and c._name not in cloud_filter:
```
tzumainn, line 50: cloud_filter has a default value, so no need for the first part of the and clause

esiclient/tests/unit/v1/test_node_baremetal.py
```diff
@@ -0,0 +1,129 @@
...
+            ]
+        )
+        self.assertEqual(expected, results)
+        mock_config.assert_called_once()
+        assert mock_cloud.list_machines.call_count == 3
+
+    @mock.patch('esiclient.v1.node_baremetal.openstack.connect')
+    @mock.patch('esiclient.v1.node_baremetal.openstack.config.loader.'
+                'OpenStackConfig.get_all_clouds')
+    def test_take_action_list(self, mock_config, mock_connect):
```
tzumainn, line 100: This test has the same name as the other one; they should have different names.

tzumainn (commented):
A few more comments; also the pep8 check is still failing.

esiclient/v1/node_baremetal.py
ljmcgann: its default value is to be empty if you dont pass in the --clouds flag. This means that nothing will print

esiclient/tests/unit/v1/test_node_baremetal.py
ljmcgann: my bad, fixed

ljmcgann:
addressed comments

tzumainn:
I forgot one more rather big change; can you rename the new file mdc_node_baremetal.py and put it into a new directory - esiclient/v1/mdc/? It seems like it'll be useful to separate these out.

ljmcgann:
moved the appropriate files to mdc directory

setup.cfg
```diff
@@ -31,6 +31,7 @@ openstack.cli.extension =
     esiclient = esiclient.plugin
 
 openstack.esiclient.v1 =
+    esi_mdc_baremetal_node_list = esiclient.v1.mdc.mdc_node_baremetal.py:MDCBaremetalNodeList
```
tzumainn, line 34: The ".py" means this doesn't work

tzumainn (commented):
I see one more pep8 error:

./esiclient/tests/unit/v1/mdc/test_mdc_node_baremetal.py:27:9: H238  old style class declaration, use new style (inherit from `object`)
        class FakeConfig():

setup.cfg
ljmcgann: yup, fixed

tzumainn (approved):

tzumainn:
Looks good - thanks!

PR #26, opened 2022-01-28
Set auth_required to False for mdc command

tzumainn:
The mdc CLI commands automatically use clouds.yaml; thus there
is no need to require any authentication.

### DIFF
--- a/esiclient/v1/mdc/mdc_node_baremetal.py
+++ b/esiclient/v1/mdc/mdc_node_baremetal.py
@@ -21,6 +21,7 @@ class MDCBaremetalNodeList(command.Lister):
     """List baremetal nodes from multiple OpenStack instances"""
 
     log = logging.getLogger(__name__ + ".List")
+    auth_required = False
 
     def get_parser(self, prog_name):
         parser = super(MDCBaremetalNodeList, self).get_parser(prog_name)


esiclient/v1/mdc/mdc_node_baremetal.py
```diff
@@ -21,6 +21,7 @@ class MDCBaremetalNodeList(command.Lister):
     """List baremetal nodes from multiple OpenStack instances"""
 
     log = logging.getLogger(__name__ + ".List")
+    auth_required = False
```
ljmcgann, line 24: How does setting the variable here even do anything?

ljmcgann:
This doesn't raise any vulnerabilities for other commands does it? I'd imagine it turns off auth for all other esi commands once this command is run once.

esiclient/v1/mdc/mdc_node_baremetal.py
tzumainn: It's the same as setting the log variable above it; it just sets a parameter which the parent class uses.

tzumainn:
> This doesn't raise any vulnerabilities for other commands does it? I'd imagine it turns off auth for all other esi commands once this command is run once.

Nope! The other commands will still look for authentication. This only applies to this command.

ljmcgann:
if thats really all there is too it then so be it!

ljmcgann (approved):

PR #31, opened 2022-05-25
Update CI to use GitHub Actions

tzumainn:


.github/workflows
```diff
@@ -0,0 +1,28 @@
...
+      - uses: actions/checkout@v3
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v3
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install tox
+          pip install -r requirements.txt
```
larsks, line 23: I don't think you need to install requirements here: `tox` builds virtual environments and then installs the requirements inside there before running the tests (https://github.com/CCI-MOC/python-esiclient/blob/master/tox.ini#L18).

larsks (approved):

PR #32, opened 2022-05-25
lint fix

tzumainn:


larsks (approved):
Sorry, I should have caught the path issue in my earlier review :)

PR #35, opened 2022-06-10
Adding License to repository in order to submit to PyPI

rd4398:
This pull request consists of a bunch files that are being added in order to build and submit package to PyPI.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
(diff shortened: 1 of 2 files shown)


tzumainn (changes requested):
Hi! I've tested this and it works well! A few questions and comments:

a) I don't think we need to include MANIFEST.in or the dist/ directory; we don't want to need these packages if we're uploading them to pypi.
b) I don't think pyproject.toml is necessary
c) Are the changes to setup.cfg needed? We already have requirements.txt and test-requirements.txt; it seems like duplication.

It may be worth trying to build these packages without these files to see if it works!

rd4398:
Okay, I will try to build these packages without the mentioned files and get back to you on how it turns out!

tzumainn (approved):

tzumainn:
Looks good - thanks!

PR #36, opened 2022-06-14
Make esi node network list more efficient

DanNiESh:


esiclient/v1/node_network.py
```diff
@@ -54,23 +54,46 @@ def take_action(self, parsed_args):
...
+                for np in neutron_ports:
+                    if np.id == neutron_port_id:
+                        neutron_port = np
+                        break
                 network_id = neutron_port.network_id
+
+                network_name = ''
+                for network in networks:
+                    if network.id == network_id:
+                        network_name = network.name
```
tzumainn, line 83: This should probably use utils.get_network_display_name

esiclient/v1/node_network.py
```diff
@@ -54,23 +54,46 @@ def take_action(self, parsed_args):
         if parsed_args.network:
             filter_network = neutron_client.find_network(parsed_args.network)
 
+        nodes = ironic_client.node.list()
+        neutron_ports = neutron_client.ports()
+        networks = neutron_client.networks()
```
tzumainn, line 59: I think these can be reduced; if parsed_args.node is defined, then we only need to fetch one node; similarly if parsed_args.network is defined. Neutron ports can also be filtered by networks if parsed_args.network is defined.

esiclient/v1/node_network.py
```diff
@@ -54,23 +54,46 @@ def take_action(self, parsed_args):
...
         data = []
+        node_name = ''
         for port in ports:
-            node = ironic_client.node.get(port.node_uuid)
+            if not parsed_args.node or not node_name:
+                for node in nodes:
+                    if node.uuid == port.node_uuid:
+                        node_name = node.name
+                        break
+
```
tzumainn, line 69: I think you can reduce the lines by searching the list of dicts as described here: https://stackoverflow.com/questions/8653516/python-list-of-dictionaries-search

tzumainn (changes requested):
Hi! This is pretty good; things work and the query time is *much* reduced. A few comments inline.

esiclient/v1/node_network.py
DanNiESh: Fixed.

DanNiESh:
> I think this is close! The last suggestion I'd make is fairly minor: in any conditional where you check `if parsed_args.network`, I'd suggest changing to `if filter_network` instead. That makes it clearer what the nature of the conditional is.

Not sure if line #100 "elif not parsed_args.network" should be changed to "filter_network". I think it should remain as it it.

tzumainn (approved):
Looks good - thanks!

PR #40, opened 2022-06-27
Clarify install commands in README

rd4398:


tzumainn (changes requested):
See comments on https://github.com/CCI-MOC/esi-leap/pull/99; in this case, maybe also add the instructions for installing from source, just for documentation parity.

rd4398:
Made changes as per your feedback!
Do let me know if any other changes are required.

tzumainn (approved):
Looks good!

PR #41, opened 2022-07-14
Fix node network list command

tzumainn:
The neutron port command retrieves a generator; we need it to be a list.


DanNiESh (approved):
LGTM

PR #42, opened 2022-08-09
'openstack node network list' command bug fix

rd4398:
I added a condition to check whether neutron port exists (on the line it was failing).
Not sure though whether this fixes the issue...
Let me know if any more changes are required!

### DIFF
--- a/esiclient/tests/unit/v1/test_node_network.py
+++ b/esiclient/tests/unit/v1/test_node_network.py
@@ -44,6 +44,12 @@ class TestList(base.TestCommand):
             "address": "cc:cc:cc:cc:cc:cc",
             "internal_info": {'tenant_vif_port_id': 'neutron_port_uuid_2'}
         })
+        self.port4 = utils.create_mock_object({
+            "uuid": "port_uuid_4",
+            "node_uuid": "11111111-2222-3333-4444-bbbbbbbbbbbb",
+            "address": "dd:dd:dd:dd:dd:dd",
+            "internal_info": {'tenant_vif_port_id': 'neutron_port_uuid_4'}
+        })
         self.node1 = utils.create_mock_object({
             "uuid": "11111111-2222-3333-4444-aaaaaaaaaaaa",
             "name": "node1"
@@ -108,12 +114,28 @@ class TestList(base.TestCommand):
 
         results = self.cmd.take_action(parsed_args)
         expected = (
+          ["Node", "MAC Address", "Port", "Network", "Fixed IP"],
+          [['node1', 'aa:aa:aa:aa:aa:aa', 'node1', 'test_network', '1.1.1.1'],
+           ['node2', 'bb:bb:bb:bb:bb:bb', None, None, None],
+           ['node2', 'cc:cc:cc:cc:cc:cc', 'node2', 'test_network', '2.2.2.2']]
+        )
+        self.assertEqual(expected, results)
+        self.app.client_manager.baremetal.port.list.\
+            assert_called_once_with(detail=True)
+
+    def test_take_action_neutron_port_does_not_exist(self):
+        self.app.client_manager.baremetal.port.list.\
+            return_value = [self.port4]
+
+        arglist = []
+        verifylist = []
+
+        parsed_args = self.check_parser(self.cmd, arglist, verifylist)
+
+        results = self.cmd.take_action(parsed_args)
+        expected = (
             ["Node", "MAC Address", "Port", "Network", "Fixed IP"],
-            [["node1", "aa:aa:aa:aa:aa:aa",
-              "node1", "test_network", "1.1.1.1"],
-             ["node2", "bb:bb:bb:bb:bb:bb", None, None, None],
-             ["node2", "cc:cc:cc:cc:cc:cc",
-              "node2", "test_network", "2.2.2.2"]]
+            [["node2", "dd:dd:dd:dd:dd:dd", None, None, None]]
         )
         self.assertEqual(expected, results)
         self.app.client_manager.baremetal.port.list.\
@@ -130,10 +152,9 @@ class TestList(base.TestCommand):
 
         results = self.cmd.take_action(parsed_args)
         expected = (
-            ["Node", "MAC Address", "Port", "Network", "Fixed IP"],
-            [["node2", "bb:bb:bb:bb:bb:bb", None, None, None],
-             ["node2", "cc:cc:cc:cc:cc:cc",
-              "node2", "test_network", "2.2.2.2"]]
+           ["Node", "MAC Address", "Port", "Network", "Fixed IP"],
+           [['node2', 'bb:bb:bb:bb:bb:bb', None, None, None],
+            ['node2', 'cc:cc:cc:cc:cc:cc', 'node2', 'test_network', '2.2.2.2']]
         )
         self.assertEqual(expected, results)
         self.app.client_manager.baremetal.port.list.\
--- a/esiclient/v1/node_network.py
+++ b/esiclient/v1/node_network.py
@@ -74,30 +74,31 @@ class List(command.Lister):
                                   if node.uuid == port.node_uuid), None).name
 
             neutron_port_id = port.internal_info.get('tenant_vif_port_id')
+            neutron_port = None
 
             if neutron_port_id:
                 neutron_port = next((np for np in neutron_ports
                                      if np.id == neutron_port_id), None)
 
+            if neutron_port is not None:
                 network_id = neutron_port.network_id
 
-                if filter_network:
-                    network_name = utils.get_network_display_name(
-                        filter_network)
-                else:
-                    network = next((network for network in networks
-                                    if network.id == network_id), None)
-                    network_name = utils.get_network_display_name(network)
-
                 if not filter_network or filter_network.id == network_id:
+                    if filter_network:
+                        network = filter_network
+                    else:
+                        network = next((network for network in networks
+                                        if network.id == network_id), None)
+
+                    network_name = utils.get_network_display_name(network)
                     fixed_ip = ''
                     if neutron_port.fixed_ips and \
                             len(neutron_port.fixed_ips) > 0:
                         fixed_ip = neutron_port.fixed_ips[0]['ip_address']
-                    data.append([node_name, port.address,
-                                 neutron_port.name,
-                                 network_name,
-                                 fixed_ip])
+                        data.append([node_name, port.address,
+                                    neutron_port.name,
+                                    network_name,
+                                    fixed_ip])
             elif not filter_network:
                 data.append([node_name, port.address, None, None, None])
 


tzumainn:
Could you add a unit test that tests this branch of the code? Thanks!

rd4398:
Sure, will try!

rd4398:
I am a bit lost about how I can write unit test for that branch of code..I tried something but got errors..
Any suggestions?

tzumainn:
Unit tests can be pretty complicated - can you push your attempt to this PR? I can make comments from there!

rd4398:
Sure, will do!

rd4398:
I pushed my attempt!

esiclient/tests/unit/v1/test_node_network.py
```diff
@@ -119,6 +119,26 @@ def test_take_action(self):
         self.app.client_manager.baremetal.port.list.\
             assert_called_once_with(detail=True)
 
+    def test_take_action_neutron_port(self):
```
tzumainn, line 122: You'll want to rename this to indicate this is a negative test case where the baremetal port indicates a matching neutron port that does not exist

esiclient/tests/unit/v1/test_node_network.py
```diff
@@ -119,6 +119,26 @@ def test_take_action(self):
...
+        self.app.client_manager.baremetal.port.list.\
+            return_value = [self.neutron_port1]
```
tzumainn, line 124: Confusingly, there are actually two different kinds of ports: baremetal ports and neutron ports. We want this to return a baremetal port - specifically one that tests the new conditional in your code, where the baremetal port has a tenant_vif_port_id that does not match an existing neutron port. So you'll want to create a "self.port4" above that does so, and use that here.

tzumainn (changes requested):
Hi! Here's some early feedback. Updating the test in the way I described will show that the changed code won't quiiiiiite work - try and figure out why, but let me know if you need help!

rd4398:
Okay, I will take a look!
Thanks

rd4398:
I did changes to the test as per your feedback. You were right, the changes in code don't actually work...
I am now a bit lost about how and where I can make the change..
Can you please help me figuring that out?

tzumainn:
Don't undo the change; it's the first step that needs to be done. Put it back in, run the test, and paste the output when it fails, and I can help you from there!

rd4398:
Okay, will do!

rd4398:
`Failed 1 tests - output below:
==============================

esiclient.tests.unit.v1.test_node_network.TestList.test_take_action_neutron_port
--------------------------------------------------------------------------------

Captured traceback:
~~~~~~~~~~~~~~~~~~~
    Traceback (most recent call last):

      File "/home/rohand/python-esiclient/esiclient/tests/unit/v1/test_node_network.py", line 137, in test_take_action_neutron_port
    results = self.cmd.take_action(parsed_args)

      File "/home/rohand/python-esiclient/esiclient/v1/node_network.py", line 89, in take_action
    network = next((network for network in networks

      File "/home/rohand/python-esiclient/esiclient/v1/node_network.py", line 90, in <genexpr>
    if network.id == network_id), None)

    NameError: free variable 'network_id' referenced before assignment in enclosing scope`

This the error I get when I run the test

esiclient/v1/node_network.py
```diff
@@ -79,7 +79,8 @@ def take_action(self, parsed_args):
                 neutron_port = next((np for np in neutron_ports
                                      if np.id == neutron_port_id), None)
 
-                network_id = neutron_port.network_id
+                if neutron_port:
+                    network_id = neutron_port.network_id
```
tzumainn, line 83: This block of code inside the "if neutron_port_id" in line 78 is essentially saying "if the network exists, find network information"; otherwise it goes to line 102/103 and puts in a bunch of 'None' instead of network information.

What's happening is that we've discovered a new case when the network secretly doesn't exist - when the assumed neutron_port also doesn't exist. The 'if' statement you've added checks that, but it doesn't redirect the code into the 'no network' case. That's why eventually it fails at line 90; that line still assumes a network exists... and it doesn't.

rd4398: Okay... I have a much clearer picture now...
I tried to redirect the code to no network case from the if statement that I added...however, it's resulting into more errors and failure of unit tests....any hint how I can do that?

rd4398:
I don't know whats wrong with me today but for some reason I am struggling with this :(
Can you please take a look?
Thanks

tzumainn (changes requested):
Hi! So the fundamental issue can be seen through the comment that you added, which is correct:

"checks whether network exists and find its information"

Right now the conditional does the following:

```
    if we expect a port:
           find network
           set network information
    else
           set no network information
```

The problem is that the 'find network' code has nowhere to go if it doesn't find a network. What you have to do is move that code outside of the conditional.

esiclient/v1/node_network.py
```diff
@@ -75,12 +75,16 @@ def take_action(self, parsed_args):
 
             neutron_port_id = port.internal_info.get('tenant_vif_port_id')
 
+#           checks whether network exists and find its information
```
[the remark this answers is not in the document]
rd4398: Okay, so the code from line 80 to 94 has to be brought out of conditional, right?
Also, I will change spacing of the comments!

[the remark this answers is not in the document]
tzumainn: Some of it does, yep!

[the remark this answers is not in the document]
rd4398: Okay, will check!

rd4398:
I shifted some of the code outside the conditional...( Not sure whether I have done it correctly ) I was wondering whether we can write an "else" to the "if" on line 82 as I am getting this error:

`Captured traceback:
~~~~~~~~~~~~~~~~~~~
    Traceback (most recent call last):

      File "/home/rohand/python-esiclient/esiclient/tests/unit/v1/test_node_network.py", line 137, in test_take_action_neutron_port
    results = self.cmd.take_action(parsed_args)

      File "/home/rohand/python-esiclient/esiclient/v1/node_network.py", line 90, in take_action
    network = next((network for network in networks

      File "/home/rohand/python-esiclient/esiclient/v1/node_network.py", line 91, in <genexpr>
    if network.id == network_id), None)

    NameError: free variable 'network_id' referenced before assignment in enclosing scope`

tzumainn:
You probably need to set a default value of 'network_id = None'

rd4398:
Yeah, I tried that...
I still get the following:

`Captured traceback:
~~~~~~~~~~~~~~~~~~~
    Traceback (most recent call last):

      File "/home/rohand/python-esiclient/esiclient/utils.py", line 21, in get_network_display_name
    network.name, network.provider_segmentation_id)

    AttributeError: 'NoneType' object has no attribute 'name'

    
During handling of the above exception, another exception occurred:


    Traceback (most recent call last):

      File "/home/rohand/python-esiclient/esiclient/tests/unit/v1/test_node_network.py", line 157, in test_take_action_node_filter
    results = self.cmd.take_action(parsed_args)

      File "/home/rohand/python-esiclient/esiclient/v1/node_network.py", line 94, in take_action
    network_name = utils.get_network_display_name(network)

      File "/home/rohand/python-esiclient/esiclient/utils.py", line 25, in get_network_display_name
    return network.name

    AttributeError: 'NoneType' object has no attribute 'name'
`

tzumainn:
The way I'd think about it is like this. In the current code, this is the "if" statement that asks if a network exists:

https://github.com/CCI-MOC/python-esiclient/blob/master/esiclient/v1/node_network.py#L78

And this is the else that handles the case if it doesn't:

https://github.com/CCI-MOC/python-esiclient/blob/master/esiclient/v1/node_network.py#L101

Now, the if statement is kind of just wrong; it *assumes* a network exists if a neutron_port_id exists, but the bug basically says that's not true; the network could still not exist.

What we want is for the `if neutron_port_id:` to begin a conditional that tries and locate the network. Once that's done, we have:

```
if filter_network:
   <do stuff>
else if network:
    <do other stuff>
else:
    <no network case>
```

rd4398:
Okay, will check!

rd4398:
Hi!
I tried to make changes as per our discussion during scrum call today
I keep getting the following error:

`Captured traceback:
~~~~~~~~~~~~~~~~~~~
    Traceback (most recent call last):

      File "/home/rohand/python-esiclient/esiclient/tests/unit/v1/test_node_network.py", line 137, in test_take_action_neutron_port
    results = self.cmd.take_action(parsed_args)

      File "/home/rohand/python-esiclient/esiclient/v1/node_network.py", line 83, in take_action
    network_id = neutron_port.network_id

    AttributeError: 'NoneType' object has no attribute 'network_id'`

I am not able to figure out the reason for this conditional on line 92 of this commit `if network.id == network_id` which also exists in original code...
Any suggestions?

esiclient/v1/node_network.py
```diff
@@ -74,31 +74,34 @@ def take_action(self, parsed_args):
                                   if node.uuid == port.node_uuid), None).name
 
             neutron_port_id = port.internal_info.get('tenant_vif_port_id')
+            network = None
 
             if neutron_port_id:
                 neutron_port = next((np for np in neutron_ports
                                      if np.id == neutron_port_id), None)
 
                 network_id = neutron_port.network_id
```
tzumainn, line 83: Ah, the reason goes back to the original reason for this PR; there are cases where the neutron_port_id is specified, but the neutron_port doesn't exist. In such cases, the neutron_port is set to None, and we get this exact error. So now you just have to add a check here to only continue if neutron_port is not None.

rd4398: Ah, okay!

rd4398:
Okay, the tests are passing now...
However not sure whether this is what is expected (since I had to change few unit tests)
Let me know if any more changes are required!

esiclient/tests/unit/v1/test_node_network.py
tzumainn: Make sure to rename this test!

esiclient/tests/unit/v1/test_node_network.py
```diff
@@ -106,14 +112,32 @@ def test_take_action(self):
...
 
+        results = self.cmd.take_action(parsed_args)
+        expected = (
+          ["Node", "MAC Address", "Port", "Network", "Fixed IP"],
+          [['node1', 'aa:aa:aa:aa:aa:aa', 'node1', 'test_network', '1.1.1.1'],
+           ['node1', 'aa:aa:aa:aa:aa:aa', None, None, None],
+           ['node2', 'bb:bb:bb:bb:bb:bb', None, None, None],
+           ['node2', 'cc:cc:cc:cc:cc:cc', 'node2', 'test_network', '2.2.2.2'],
+           ['node2', 'cc:cc:cc:cc:cc:cc', None, None, None]]
+        )
```
tzumainn, line 123: This is definitely not expected; now we're getting the same mac address returned multiple times, once with the network fields filled out and once without. That means there's something wrong with the updated code.

tzumainn (changes requested):
It's close! Try this:

```
        for port in ports:
            if not parsed_args.node:
                node_name = next((node for node in nodes
                                  if node.uuid == port.node_uuid), None).name

            neutron_port_id = port.internal_info.get('tenant_vif_port_id')
            if neutron_port_id:
                neutron_port = next((np for np in neutron_ports
                                     if np.id == neutron_port_id), None)

            if neutron_port is not None:
                network_id = neutron_port.network_id

                if not filter_network or filter_network.id == network_id:
                    if filter_network:
                        network = filter_network
                    else:
                        network = next((network for network in networks
                                        if network.id == network_id), None)
                    network_name = utils.get_network_display_name(network)
                    fixed_ip = ''
                    if neutron_port.fixed_ips and \
                            len(neutron_port.fixed_ips) > 0:
                        fixed_ip = neutron_port.fixed_ips[0]['ip_address']
                        data.append([node_name, port.address,
                                    neutron_port.name,
                                    network_name,
                                    fixed_ip])
            elif not filter_network:
                data.append([node_name, port.address, None, None, None])
```

rd4398:
Okay will try!
Can you suggest a name for the test?
Don't really know how to name negative test case...

tzumainn:
Let's try something like 'test_take_action_neutron_port_does_not_exist'

rd4398:
The above snippet worked!

I just had to initialize `neutron_port = None`

Let me know if any more changes are required!

tzumainn (approved):
Looks good - thanks!

PR #43, opened 2022-09-02
Add --trunk option to node network list

DanNiESh:


esiclient/utils.py
```diff
@@ -39,7 +39,7 @@ def get_network_info_from_port(port, client):
     return get_network_display_name(network), fixed_ip
 
 
-def get_full_network_info_from_port(port, client):
+def get_full_network_info_from_port(port, client, get_trunk_details=True):
```
tzumainn, line 42: I think this is better defaulted as "False", as that's the more usual default case.

tzumainn (changes requested):
Looks really good! One minor comment inline.

esiclient/utils.py
DanNiESh: Fixed.

esiclient/utils.py
```diff
@@ -39,7 +39,7 @@ def get_network_info_from_port(port, client):
     return get_network_display_name(network), fixed_ip
 
 
-def get_full_network_info_from_port(port, client):
+def get_full_network_info_from_port(port, client, get_trunk_details=False):
```
tzumainn, line 42: Apologies - I thought the get_trunk_details switch was called from the same function, but they're not. That makes me think that this would actually be cleaner split into two functions, with the trunk details calling this one.

DanNiESh: Fixed! Also tested on pilot deployment with command "openstack esi node network attach oct4-10 --trunk naved-created-trunk". It successfully attached esi-naved-created-trunk-provisioning-trunk-port to oct4-10.

DanNiESh:
updated output is:
+-------------+-------------------------------------------------+
| Field       | Value                                           |
+-------------+-------------------------------------------------+
| Node        | oct4-10                                         |
| MAC Address | 00:0f:53:57:83:71                               |
| Port        | esi-naved-created-trunk-provisioning-trunk-port |
|             | esi-naved-created-trunk-naved-personal-sub-port |
| Network     | provisioning (623)                              |
|             | naved-personal (625)                            |
| Fixed IP    | 192.168.11.139                                  |
|             | 192.168.12.157                                  |
+-------------+-------------------------------------------------+

tzumainn (approved):
Looks good - thanks!

PR #44, opened 2022-09-07
Add ESI switch CLI commands

tzumainn:
* list vlans
* enable access to switchport
* disable access to switchport


README.md
```diff
@@ -136,3 +136,38 @@ openstack esi node volume attach
...
+```
+### `openstack esi switch port enable access`
+
+Attach VLAN to a switchport.
+
+```
+openstack esi switch port enable access
+   <switchport>
+   <vlan>
+```
```
naved001, line 159: A general question about the command line: How do we specify the switch whose switchport we want to operate on? Also this switchport is the real switchport, right? like `Ethernet1/3`?

naved001: When I say "whose switchport" I mean we will generally have more than one switch. So `Ethernet1/3` will be on a switch e.g. `cisco-1` and `cisco-2`.

tzumainn: Ah, I'm assuming that each ESI only deals with a single switch right now - is that likely to change?

Switchport is the real switchport, correct!

tzumainn: > When I say "whose switchport" I mean we will generally have more than one switch. So `Ethernet1/3` will be on a switch e.g. `cisco-1` and `cisco-2`.

Oh - it would be whichever switch ESI is configured to work with.

naved001: > I'm assuming that each ESI only deals with a single switch right now - is that likely to change?

When we add port information to a baremetal node it does take in all the switch information (switch IP etc), so I assumed that for a separate node we could put a different switch. I was under the assumption that multiple switches are supported out of the box. Sorry if I missed this somewhere.

tzumainn: > > I'm assuming that each ESI only deals with a single switch right now - is that likely to change?
> 
> When we add port information to a baremetal node it does take in all the switch information (switch IP etc), so I assumed that for a separate node we could put a different switch. I was under the assumption that multiple switches are supported out of the box. Sorry if I missed this somewhere.

Oh, no worries! The important specification is actually a switch identifier, which corresponds to a configuration in ml2_conf.ini. We could in theory have multiple switch configurations in ml2_conf.ini, and different nodes could specify different switches. Is that likely? I had thought that would actually be different instances of ESI.

naved001: > Is that likely?

Yes, in fact, it is extremely unlikely that an ESI cluster will only manage a single switch. The expectation is that it will be used to manage multiple switches (only the switchports where the nodes connect to, not the interswitch links) and those switches may also be of different kind.

tzumainn: Ah, okay! I'll add a required switch parameter.

tzumainn: @naved001 okay, I've updated the commands - let me know what you think!

esiclient/v1/switch.py
```diff
@@ -0,0 +1,174 @@
...
+            provider_segmentation_id=vlan_id))
+        network = next(iter(networks), None)
+        if not network:
+            raise exceptions.CommandError("ERROR: VLAN ID unknown")
+
+        # attach node to network
+        vif_info = {}
+        np_name = utils.get_port_name(network.name, prefix=node.name)
+        np = utils.get_or_create_port(np_name, network, neutron_client)
+        ironic_client.node.vif_attach(node.uuid, np.id, **vif_info)
```
DanNiESh, line 126: I saw vif_info elsewhere in esi code. What message should it contain and why we use an empty dict?

tzumainn: It's used (and populated) when attaching a node to a network by mac address; I think it's a required field is why.

esiclient/v1/switch.py
```diff
@@ -0,0 +1,174 @@
...
+        self.log.debug("take_action(%s)", parsed_args)
+
+        switch = parsed_args.switch
+
+        ironic_client = self.app.client_manager.baremetal
+        neutron_client = self.app.client_manager.network
+
+        ports = list((port for port in ironic_client.port.list(detail=True)
+                      if port.local_link_connection.get(
+                              'switch_info') == switch))
```
naved001, line 45: I am trying to understand what happens when the switch does not exist. I see in the tests you are testing for scenarios with an unknown switch, but the error those expect is "Switch port unknown", I would expect it to be more clear that the switch doesn't exist.

naved001 (commented):
looks good, just one question.

esiclient/v1/switch.py
tzumainn: Ah, that's actually extremely difficult to do, since OpenStack doesn't have a separate entity for switches; it's just information in a baremetal port. To be honest I don't hate obfuscating the exact kind of failure if someone is just randomly throwing out switch and switchports.

naved001: > Ah, that's actually extremely difficult to do, since OpenStack doesn't have a separate entity for switches;

Fair enough.

> is just randomly throwing out switch and switchports.

I was thinking more of a situation where someone made a typo in entering the switch name but then the CLI tells you the switchport can't be found and that may be confusing. But if there's not an easy way to do it then we could ignore it.

naved001 (approved):

PR #45, opened 2022-09-09
Fix issue when attaching a node to a network results in no fixed ip

tzumainn:
There was a minor bug introduced recently; a misaligned indent meant that if a port did not have a fixed IP, it would not be represented in the node network list. This fixes that issue.


DanNiESh (approved):
Thanks!

PR #46, opened 2022-09-13
Add switch trunk commands

tzumainn:
* add vlan to switchport as trunk
* add vlan to trunk
* remove vlan from trunk


tzumainn:
@naved001 hi! I'm still writing tests for this, but I was hoping for some early feedback. Does this workflow look right?

```
   openstack esi switch port enable access trunk       # add vlan to switch port as trunk
   openstack esi switch trunk add vlan                        # add tagged vlan to trunk
   openstack esi switch trunk remove vlan                  # remove tagged vlan from trunk
   openstack esi switch port disable access                # remove trunk from switch port
```

naved001:
@tzumainn I would remove the word "access" from those commands, since that's also used to describe the switchport mode (access mode vs trunk mode). 

Also, what are the arguments for those commands? Here's what I am thinking.

1. `openstack esi switch port enable trunk <switch> <switchport> <native-vlan>`. This should set the switchport in trunk mode and set the native-vlan (same as untagged vlan). 
 
2. `openstack esi switch trunk add vlan <switch> <switchport> <tagged-vlan-to-add>`. This should add additional tagged vlans.

3. `openstack esi switch trunk remove vlan <switch> <switchport> <tagged-vlan-to-remove>`. This should remove the tagged vlan. Note: this should not remove the native-vlan because some switches end up setting VLAN 1 as the native-vlan if no native-vlan is configured in trunk mode, which is bad. If someone wanted to set a different native-vlan, they should disable trunk and then start again. That would be a safe way of doing things without making the code unnecessarily complex and too dependent on the type of switch.

4. `openstack esi switch port disable trunk`. This should remove all tagged VLANs, the untagged native-vlan, and turn off the switchport.

Let me know what you think

tzumainn:
@naved001 makes sense! The arguments are pretty much what you described. I'll update the code - thanks!

esiclient/v1/switch.py
```diff
@@ -172,3 +171,284 @@ def take_action(self, parsed_args):
...
+        neutron_client = self.app.client_manager.network
+        networks = list(neutron_client.networks(
+            provider_network_type='vlan',
+            provider_segmentation_id=vlan_id))
+        network = next(iter(networks), None)
+        if not network:
+            raise exceptions.CommandError("ERROR: VLAN ID unknown")
+
+        # create trunk
+        trunk_name = switch + switchport
```
DanNiESh, line 228: should it be switch +  '-'  + switchport?

esiclient/v1/switch.py
```diff
@@ -172,3 +171,284 @@ def take_action(self, parsed_args):
...
+    def take_action(self, parsed_args):
+        self.log.debug("take_action(%s)", parsed_args)
+
+        switch = parsed_args.switch
+        switchport = parsed_args.switchport
+        vlan_id = parsed_args.vlan_id
+
+        # get associated port and node
+        ironic_client = self.app.client_manager.baremetal
+        ports = ironic_client.port.list(detail=True)
```
DanNiESh, line 207: I think this code block is used a lot in different functions (line 207 - line 213). Should we create a function like utils.get_barmetal_port() to avoid duplication?

esiclient/v1/switch.py
```diff
@@ -172,3 +171,284 @@ def take_action(self, parsed_args):
...
+            [
+                {
+                    'port_id': sub_port.id,
+                    'segmentation_type': 'vlan',
+                    'segmentation_id': vlan_id
+                }
+            ]
+        )
+
+        return ["Switchport", "VLAN", "Node", "Network"], \
```
DanNiESh, line 319: Same.

esiclient/v1/switch.py
```diff
@@ -172,3 +171,284 @@ def take_action(self, parsed_args):
...
+        trunk = neutron_client.delete_trunk_subports(
+            trunk.id,
+            [
+                {
+                    'port_id': sub_port.id,
+                }
+            ]
+        )
+
+        return ["Switchport", "VLAN", "Node", "Network"], \
```
DanNiESh, line 398: Same. Trunk name.

esiclient/v1/switch.py
```diff
@@ -172,3 +171,284 @@ def take_action(self, parsed_args):
...
+            trunk_port_name, network, neutron_client)
+        neutron_client.create_trunk(
+            name=trunk_name,
+            port_id=trunk_port.id,
+        )
+
+        # attach node to network
+        ironic_client.node.vif_attach(node.uuid, trunk_port.id)
+
+        return ["Switchport", "VLAN", "Node", "Network"], \
```
DanNiESh, line 241: Do we want to add trunk name to the return value?

esiclient/v1/switch.py
tzumainn: Good point! Fixed here (and other places)

esiclient/v1/switch.py
tzumainn: Fixed this (and others) as well!

DanNiESh (approved):
Looks good!

PR #51, opened 2022-10-14
Add floating ip info in esi node network list

DanNiESh:
1. Expand 'openstack esi node network list' output to  Node|MAC Address|Port |Network|Fixed IP|Floating IP|Floating Network
2. Reduce neutron /networks/id API calls by fetching all networks at once and then convert the network list into a dict {network_id: network_obj}.

### DIFF
--- a/esiclient/utils.py
+++ b/esiclient/utils.py
@@ -25,13 +25,17 @@ def get_network_display_name(network):
         return network.name
 
 
-def get_network_info_from_port(port, client):
+def get_network_info_from_port(port, client, networks_dict={}):
     """Return Neutron network name and ips from port
 
     :param port: a Neutron port
     :param client: neutron client
+    :param networks_dict: networks dict {id:network}
     """
-    network = client.get_network(port.network_id)
+    if port.network_id in networks_dict:
+        network = networks_dict.get(port.network_id)
+    else:
+        network = client.get_network(port.network_id)
     fixed_ip = ''
     if port.fixed_ips and len(port.fixed_ips) > 0:
         fixed_ip = port.fixed_ips[0]['ip_address']
@@ -39,19 +43,21 @@ def get_network_info_from_port(port, client):
     return get_network_display_name(network), fixed_ip
 
 
-def get_full_network_info_from_port(port, client):
+def get_full_network_info_from_port(port, client, networks_dict={}):
     """Return full Neutron network name and ips from port
 
     This code iterates through subports if appropriate
 
     :param port: a Neutron port
     :param client: neutron client
+    :param networks_dict: networks dict {id:network}
     """
     network_names = []
     port_names = []
     fixed_ips = []
 
-    network_name, fixed_ip = get_network_info_from_port(port, client)
+    network_name, fixed_ip = get_network_info_from_port(port, client,
+                                                        networks_dict)
     network_names.append(network_name)
     fixed_ips.append(fixed_ip)
     port_names.append(port.name)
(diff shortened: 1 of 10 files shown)


DanNiESh:
I've made the change to reduce the get_network API calls. Let me know what do you think about it.

esiclient/utils.py
```diff
@@ -25,33 +25,39 @@ def get_network_display_name(network):
...
 
     :param port: a Neutron port
+    :param networks_dict: networks dict {id:network}
     :param client: neutron client
     """
-    network = client.get_network(port.network_id)
+    if networks_dict:
+        network = networks_dict.get(port.network_id)
+    else:
+        network = client.get_network(port.network_id)
```
tzumainn, line 38: I think it may be safer to have networks_dict default to {}, and to have the conditional be `if the networks_dict contains the network_id as a key`.

esiclient/utils.py
```diff
@@ -25,33 +25,39 @@ def get_network_display_name(network):
...
     fixed_ip = ''
     if port.fixed_ips and len(port.fixed_ips) > 0:
         fixed_ip = port.fixed_ips[0]['ip_address']
 
     return get_network_display_name(network), fixed_ip
 
 
-def get_full_network_info_from_port(port, client):
+def get_full_network_info_from_port(port, client, networks_dict=None):
```
tzumainn, line 46: If you take the suggestion above then networks_dict here would also default to {}

esiclient/v1/node_network.py
```diff
@@ -92,15 +97,26 @@ def take_action(self, parsed_args):
...
+                                 if floating_network_names else None]
+                                )
             elif not filter_network:
-                data.append([node_name, port.address, None, None, None])
+                data.append([node_name, port.address,
+                             None, None, None, None, None])
 
-        return ["Node", "MAC Address", "Port", "Network", "Fixed IP"], data
+        return ["Node", "MAC Address", "Port", "Network", "Fixed IP",
+                "Floating IP", "Floating Network"], data
```
tzumainn, line 119: It may make sense to switch around "Floating IP" and "Floating Network" to match "Network" and "Fixed IP"

tzumainn (commented):
Looks pretty good! A few minor comments inline.

esiclient/utils.py
DanNiESh: Done!

esiclient/v1/node_network.py
DanNiESh: Fixed!

esiclient/utils.py
```diff
@@ -25,33 +25,39 @@ def get_network_display_name(network):
...
-def get_network_info_from_port(port, client):
+def get_network_info_from_port(port, client, networks_dict={}):
     """Return Neutron network name and ips from port
 
     :param port: a Neutron port
+    :param networks_dict: networks dict {id:network}
     :param client: neutron client
     """
-    network = client.get_network(port.network_id)
+    if networks_dict and port.network_id in networks_dict:
```
tzumainn, line 35: Suuuuper minor - you no longer need the first part of this conditional, because the function will always define a networks_dict

tzumainn:
Thanks! Looks good. One last thing - I also can't figure out the issue with python 3.7. For now, can you edit .github/workflows/test.yml and remove 3.7?

tzumainn:
Thanks! Looks good. One last thing - I also can't figure out the issue with python 3.7. For now, can you edit .github/workflows/test.yml and remove 3.7?

tzumainn (approved):
Looks good!

PR #52, opened 2023-01-19
Updated CI and tox envlist to enable testing against more Python versions.

samuel-zuk:
Currently, tox tests against only Python 3.6 and 3.7 in particular. This patch changes this to "whatever Python 3 version is set up as default" so that this project can be tested in CI and in containers more easily against more versions of Python.

Additionally, as part of https://github.com/CCI-MOC/esi/issues/330, this project will now lint check and run unit tests during CI for versions of Python between 3.6, the earliest version that doesn't cause major dependency breakages-- thru 3.11, the newest version currently available by default on the Ubuntu 20.04 CI runner. (Ubuntu 20.04 is itself, currently the newest version of Ubuntu with Python 3.6 installed by default)


tzumainn (approved):
Looks good! Thanks

PR #53, opened 2023-06-28
node-network-attach-error-message

skanthed:


skanthed:
Sure, In the next commit I will fix it.

esiclient/v1/node_network.py
```diff
@@ -214,27 +214,39 @@ def take_action(self, parsed_args):
...
+                print("Attaching network {1} to node {0}{2}".format(
+                    node.name, network.name, mac_string))
+                port_name = utils.get_port_name(network.name, prefix=node.name)
+                port = utils.get_or_create_port(port_name, network,
+                                                neutron_client)
+                ironic_client.node.vif_attach(node_uuid, port.id, **vif_info)
+                port = neutron_client.get_port(port.id)
+            else:
+                print("The specified network does not exist."
+                      "Please provide a valid network name.")
```
tzumainn, line 227: Instead of printing an error message, you'll want to raise `exceptions.CommandError` (you can find examples of this if you grep in the repository). You should also check if the network exists right away - up in line 177.

esiclient/v1/node_network.py
```diff
@@ -214,27 +214,39 @@ def take_action(self, parsed_args):
...
         elif parsed_args.trunk:
             print("Attaching trunk {1} to node {0}{2}".format(
                 node.name, trunk.name, mac_string))
             port = neutron_client.get_port(trunk.port_id)
             ironic_client.node.vif_attach(node_uuid, port.id, **vif_info)
 
-        network_names, port_names, fixed_ips \
-            = utils.get_full_network_info_from_port(
-                port, neutron_client)
+        if port is not None:
```
tzumainn, line 234: This `if` won't be needed if you check for the network right away.

esiclient/v1/node_network.py
```diff
@@ -214,27 +214,39 @@ def take_action(self, parsed_args):
...
-            [node.name, port.mac_address,
-             "\n".join(port_names),
-             "\n".join(network_names),
-             "\n".join(fixed_ips)]
+            return ["Node", "MAC Address", "Port", "Network", "Fixed IP"], \
+                [node.name, port.mac_address,
+                 "\n".join(port_names),
+                 "\n".join(network_names),
+                 "\n".join(fixed_ips)]
+        else:
```
tzumainn, line 244: We don't really want to return a blank result.

tzumainn (changes requested):
Good start! Comments inline.

test-requirements.txt
```diff
@@ -8,6 +8,7 @@ docutils>=0.11 # OSI-Approved Open Source, Public Domain
 fixtures>=3.0.0 # Apache-2.0/BSD
 metalsmith
 mock>=3.0.0 # BSD
+munch>=3.0.0 # MIT
```
tzumainn, line 11: Why is this needed?

tzumainn (changes requested):
Small inline comment; the bigger item is you'll want to create a unit test covering this change. Let me know if you'd like some guidance regarding how/where to do that!

test-requirements.txt
skanthed: Was getting the error - ModuleNotFoundError: No module named 'munch' .
The specific failure occurred while trying to import the test module esiclient.tests.unit.v1.mdc.test_mdc_node_baremetal.

It seems that the test module test_mdc_node_baremetal.py has a dependency on the munch module, but it is not installed in our test environment.
In earlier PR's, they had different python versions and packages, the packages didn't depend on the munch

tzumainn: Ah, I see - good catch!

esiclient/v1/node_network.py
```diff
@@ -175,6 +175,9 @@ def take_action(self, parsed_args):
         if parsed_args.network:
             network = neutron_client.find_network(parsed_args.network)
             port = None
+            if network is None:
+                raise exceptions.CommandError(
+                    "ERROR: This is not a network name")
```
tzumainn, line 180: I'd say something like 'Unknown network' instead, since I think the user can also pass in a network UUID.

esiclient/tests/unit/v1/test_node_network.py
```diff
@@ -469,6 +473,24 @@ def test_take_action_no_port_or_network_or_trunk_exception(self):
...
+            return_value = self.node
+        self.app.client_manager.baremetal.port.list.\
+            return_value = [self.port1]
+
+        arglist = ['node1', '--network', 'test_network_invalid']
+        verifylist = []
+
+        parsed_args = self.check_parser(self.cmd, arglist, verifylist)
+#       results = self.cmd.take_action(parsed_args)
+#       print("Invalid network exception results {}".format(results))
```
tzumainn, line 487: Make sure to remove these commented out lines!

tzumainn (changes requested):

esiclient/v1/node_network.py
skanthed: I would say something like - "ERROR: This is not a network name or UUID" because then it aligns with the error message we are passing for the wrong port name or trunk name below.

esiclient/tests/unit/v1/test_node_network.py
skanthed: Sure! They are just for my test purpose.

skanthed:
@tzumainn can you look into the test case of this one? I tried multiple things but still got errors.

esiclient/v1/node_network.py
tzumainn: To be honest, I'd change it because whatever the user passes in *is* a network name; it's just not necessarily one that's in the system.

esiclient/tests/unit/v1/test_node_network.py
```diff
@@ -469,6 +473,24 @@ def test_take_action_no_port_or_network_or_trunk_exception(self):
             'ERROR: You must specify either network, port, or trunk',
             self.cmd.take_action, parsed_args)
 
+    def test_take_action_invalid_network_exception(self, mock_gfnifp):
```
tzumainn, line 476: You're missing the decorator here - actually, you probably don't need mock_gfnifp if you're not using it.

skanthed: yes I removed it, I don't think I need to use it here. It was here by mistake because I was trying something when it didn't work.

esiclient/tests/unit/v1/test_node_network.py
```diff
@@ -469,6 +473,22 @@ def test_take_action_no_port_or_network_or_trunk_exception(self):
             'ERROR: You must specify either network, port, or trunk',
             self.cmd.take_action, parsed_args)
 
+    def test_take_action_invalid_network_exception(self):
+        self.app.client_manager.baremetal.node.get.\
+            return_value = self.node
+        self.app.client_manager.baremetal.port.list.\
+            return_value = [self.port1]
+
```
tzumainn, line 481: Oh, you want to mock the call to `find_network` as well to have it explicitly return `None`.

esiclient/tests/unit/v1/test_node_network.py
```diff
@@ -469,6 +473,33 @@ def test_take_action_no_port_or_network_or_trunk_exception(self):
...
+    #             return_value=(["test_network_invalid"], ["node1-port"],
+    #                           ["2.2.2.2"]),
+    #             autospec=True)
+    def test_take_action_invalid_network_exception(self):
+        self.app.client_manager.baremetal.node.get.\
+            return_value = self.node
+        self.app.client_manager.baremetal.port.list.\
+            return_value = [self.port1]
+        self.app.client_manager.network.find_network.\
+            return_value = self.network
```
tzumainn, line 486: So what's happening here is that even though you are passing in 'test_network_invalid' - a network that doesn't exist - you are also telling `find_network` to always return a network. Because of that, this test will always think that the network exists. What you want to do is set this return_value to None instead.

esiclient/tests/unit/v1/test_node_network.py
```diff
@@ -282,6 +282,10 @@ def setUp(self):
             "id": "network_uuid",
             "name": "test_network"
         })
+        self.invalid_network = utils.create_mock_object({
+            "id": "network_uuid_invalid",
+            "name": "test_network_invalid"
+        })
```
tzumainn, line 288: I don't think this is used anywhere, so it's safe to remove.

esiclient/v1/node_network.py
```diff
@@ -174,7 +174,11 @@ def take_action(self, parsed_args):
 
         if parsed_args.network:
             network = neutron_client.find_network(parsed_args.network)
+            print("Network {} and length {}".format(network, type(network)))
```
tzumainn, line 177: This should be removed.

tzumainn (changes requested):
Close! Just a few debugging items to remove.

tzumainn:
Looks good - thanks!

PR #55, opened 2023-08-01
made port optional in "esi node network detach" command

skanthed:


esiclient/v1/node_network.py
```diff
@@ -264,21 +264,48 @@ def take_action(self, parsed_args):
         ironic_client = self.app.client_manager.baremetal
         neutron_client = self.app.client_manager.network
 
-        node = None
-        port = None
+        ports = list((port for port in ironic_client.port.list(detail=True)))
```
larsks, line 267: If you want a list, just use a list comprehension instead of using generator expression and then transforming it into a list:

```
ports = [port for port in ironic_client.port.list(detail=True)]
```

esiclient/v1/node_network.py
```diff
@@ -264,21 +264,48 @@ def take_action(self, parsed_args):
...
+        elif(len(mapped_node_port_list) == 1):
+            port_uuid = mapped_node_port_list[0].\
+                internal_info["tenant_vif_port_id"]
+
+        port = None
+
+        with concurrent.futures.ThreadPoolExecutor() as executor:
             f2 = executor.submit(neutron_client.find_port, port_uuid)
-            node = f1.result()
             port = f2.result()
```
larsks, line 301: It looks like you're submitting requests to an executor...but then you're immediately asking for a result. Unless I am missing something, you could replace these lines with:

```suggestion
        port = neutron_client.find_port(port_uuid)
```

And you would have the same behavior.

esiclient/v1/node_network.py
skanthed: okay will do this.

tzumainn, line 267: `ironic_client.port.list` should take a node parameter that limits the results to the baremetal ports associated with the node; you can see an example at https://github.com/CCI-MOC/python-esiclient/blob/master/esiclient/v1/node_network.py#L52. Doing so should remove the need to manually filter the port list below.

Also, if the user passes in a port, there's no need to do this, so you probably want to check that first - something like:

```
if parsed_args.port:
   port_uuid = parsed_args.port
else:
   <find the port_uuid by listing the baremetal ports>
```

esiclient/v1/node_network.py
```diff
@@ -264,21 +264,48 @@ def take_action(self, parsed_args):
...
+        for port in ports:
+            if(port_uuid is not None and port_uuid != ""):
+                if(port.node_uuid == node_uuid1 and port.uuid == port_uuid):
+                    mapped_node_port_list.append(port)
+            else:
+                if(port.node_uuid == node_uuid1):
+                    mapped_node_port_list.append(port)
+        if(len(mapped_node_port_list) == 0):
+            raise exceptions.CommandError(
+                "ERROR: The port is not associated with node {0}".format(
```
tzumainn, line 287: I might say "Node {0} has no associated ports"

tzumainn (changes requested):
Hi! I've left some comments. In addition, you'll need to update the README documentation for the command since the signature has changed.

esiclient/v1/node_network.py
```diff
@@ -264,21 +264,40 @@ def take_action(self, parsed_args):
...
-            node = f1.result()
-            port = f2.result()
-
-        if not port:
+        if current_node is None:
             raise exceptions.CommandError(
-                "ERROR: Port {1} not attached to node {0}".format(
-                    node.name, port_uuid))
+                "ERROR: The node {0} is not available".format(
+                    node_uuid))
```
tzumainn, line 272: I don't think you actually need this check, simply because `ironic_client.node.get` already returns an error message if the node is not available; you may want to test to be sure, but there's a reason why this check didn't exist earlier.

esiclient/v1/node_network.py
```diff
@@ -264,21 +264,40 @@ def take_action(self, parsed_args):
         ironic_client = self.app.client_manager.baremetal
         neutron_client = self.app.client_manager.network
 
-        node = None
-        port = None
+        current_node = ironic_client.node.get(node_uuid)
```
tzumainn, line 267: Not a huge deal, but I'd keep this as `node` simply because I don't see a reason to rename this variable.

esiclient/v1/node_network.py
```diff
@@ -264,21 +264,40 @@ def take_action(self, parsed_args):
...
+                    node_uuid))
+
+        if port_uuid:
+            port = neutron_client.find_port(port_uuid)
+            if not port:
+                raise exceptions.CommandError(
+                    "ERROR: Port {1} not attached to node {0}".format(
+                        current_node.name, port_uuid))
+        else:
+            ports = ironic_client.port.list(node=node_uuid, detail=True)
```
tzumainn, line 281: Can you rename this to `baremetal_ports` or `bm_ports`? I think it's important for someone else looking at the code to be able to distinguish between baremetal ports and neutron ports.

esiclient/v1/node_network.py
```diff
@@ -264,21 +264,40 @@ def take_action(self, parsed_args):
...
+
+            mapped_node_port_list = []
+            for port in ports:
```
tzumainn, line 284: This might be `for bm_port in bm_ports`

esiclient/v1/node_network.py
```diff
@@ -264,21 +264,40 @@ def take_action(self, parsed_args):
...
+                if port.internal_info.get("tenant_vif_port_id"):
+                    mapped_node_port_list.append(port)
```
tzumainn, line 286: I think you might as well append the neutron port here instead of the baremetal port; if you do, then line 296 becomes `port_uuid = mapped_port_list[0]`

esiclient/v1/node_network.py
```diff
@@ -264,21 +264,40 @@ def take_action(self, parsed_args):
...
+            if(len(mapped_node_port_list) == 0):
+                raise exceptions.CommandError(
+                    "ERROR: Node {0} is not assosiated with any port".format(
```
tzumainn, line 289: Small spelling mistake - it's `associated`

esiclient/v1/node_network.py
```diff
@@ -264,21 +264,40 @@ def take_action(self, parsed_args):
...
+                        current_node.name))
+            elif(len(mapped_node_port_list) > 1):
+                raise exceptions.CommandError(
+                    "ERROR: Multiple ports for {0} Specify the port".format(
```
tzumainn, line 293: For parallelism in error messages, I'd make this `ERROR: Node {0} is associated with multiple ports. Port must be specified with --port`

tzumainn (changes requested):
Hi! This is really close; I've left some minor comments. In addition, please remember to update the README - thanks!

skanthed:
@tzumainn If all looks good. I will update the readme too.

skanthed:
@tzumainn Updated the readme!

tzumainn (approved):
Looks good - thanks!

PR #56, opened 2023-11-06
Initial iteration of openshift orchestrate command

tzumainn:
A simple CLI command that orchestrates an OpenShift cluster based on an input config file.

### DIFF
--- a/esiclient/v1/orchestrator/openshift.py
+++ b/esiclient/v1/orchestrator/openshift.py
@@ -1,0 +1,368 @@
+#   Licensed under the Apache License, Version 2.0 (the "License"); you may
+#   not use this file except in compliance with the License. You may obtain
+#   a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#   WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#   License for the specific language governing permissions and limitations
+#   under the License.
+
+import json
+import logging
+import os
+import requests
+import time
+
+from osc_lib.command import command
+from osc_lib.i18n import _
+
+from esiclient import utils
+
+
+BASE_ASSISTED_INSTALLER_URL = \
+    'https://api.openshift.com/api/assisted-install/v2/'
+
+
+def call_assisted_installer_api(url, method, headers={}, data=None):
+    full_url = BASE_ASSISTED_INSTALLER_URL + url
+    if method == 'post':
+        response = requests.post(full_url, headers=headers, json=data)
+    elif method == 'get':
+        response = requests.get(full_url, headers=headers)
+    elif method == 'patch':
+        response = requests.patch(full_url, headers=headers, json=data)
+    else:
+        raise OSAIException(
+            'Unknown method used when calling Assisted Installer')
+    if response.status_code not in [200, 201, 202, 204]:
+        raise OSAIException(
+            "Unexpected response from Assisted Installer (%s): %s"
+            % (response.status_code, response.reason))
+    return response
+
+
+def wait_for_nodes(infra_env_id, headers, num_nodes, target_status):
+    waiting = True
+    print("waiting for hosts to reach %s..." % target_status)
+    while waiting:
+        response = call_assisted_installer_api(
+            "infra-envs/%s/hosts" % infra_env_id,
+            'get', headers)
+        count = 0
+        print("* checking hosts")
+        for host in response.json():
+            status = host.get('status')
+            print("  * " + host.get('requested_hostname') + ": " + status)
+            if status == target_status:
+                count = count + 1
+        if count == num_nodes:
+            waiting = False
+        else:
+            time.sleep(30)
+    print("... hosts ready")
+    return
+
+
+class OSAIException(Exception):
+    pass
+
+
+class OrchestrationException(Exception):
+    pass
+
+
+class Orchestrate(command.Lister):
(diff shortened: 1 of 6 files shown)


esiclient/v1/orchestrator/openshift.py
```diff
@@ -0,0 +1,339 @@
...
+        neutron_client = self.app.client_manager.network
+
+        headers = {
+            'Content-Type': 'application/json',
+            'Authorization': 'Bearer ' + os.getenv('API_TOKEN', ''),
+        }
+
+        print("STARTING OPENSHIFT CLUSTER INSTALL")
+
+        # get cluster id
```
DanNiESh, line 146: Can you break down this take_action() method into smaller methods? Such as _get_cluster_id(), _register nodes(), _install_cluster(), etc.

esiclient/v1/orchestrator/openshift.py
```diff
@@ -0,0 +1,339 @@
...
+        private_subnet_name = cluster_config.get('private_subnet_name')
+        api_vip = cluster_config.get('api_vip')
+        ingress_vip = cluster_config.get('ingress_vip')
+        openshift_version = cluster_config.get('openshift_version')
+        high_availability_mode = cluster_config.get(
+            'high_availability_mode', 'Full')
+        base_dns_domain = cluster_config.get('base_dns_domain')
+        ssh_public_key = cluster_config.get('ssh_public_key')
+
+        pull_secret = json.loads(os.environ['PULL_SECRET'])
```
DanNiESh, line 134: It might be better to check if this PULL_SECRET exists and handle the error if it doesn't exist.

DanNiESh (changes requested):

esiclient/v1/orchestrator/openshift.py
```diff
@@ -0,0 +1,339 @@
...
+        flags = ""
+        if cluster_id:
+            flags = "--cluster-id %s" % cluster_id
+            if infra_env_id:
+                flags = "%s --infra-env-id %s" % (flags, infra_env_id)
+        command = "openstack esi orchestrate openshift %s %s" % \
+            (flags, cluster_config_file)
+        if message:
+            print(message)
+        print("You may need to refresh your OpenShift token.")
```
DanNiESh, line 89: I think it might make users confused if they get an error that is not caused by token expiration, but still get the token refresh hint message. Then they just ignore the real error and rerun 'esi orchestrate openshift' anyway.

esiclient/v1/orchestrator/openshift.py
```diff
@@ -0,0 +1,339 @@
...
+
+        return parser
+
+    def take_action(self, parsed_args):
+        self.log.debug("take_action(%s)", parsed_args)
+
+        cluster_id = parsed_args.cluster_id
+        infra_env_id = parsed_args.infra_env_id
+
+        with open(parsed_args.cluster_config_file) as f:
```
DanNiESh, line 117: Do we want validate the config file? If some of the required values are missed then we can detect the errors at the beginning. Also, is it possible to add an example config file in a readme doc and state that which parameters are required and which are optional.

esiclient/v1/orchestrator/openshift.py
tzumainn: I actually tried that earlier, and it made the code really ugly, since a function such as _register_nodes will require a ton of openshift and openstack-relevant parameters.

esiclient/v1/orchestrator/openshift.py
tzumainn: Good point - fixed!

esiclient/v1/orchestrator/openshift.py
tzumainn: Added the validation! The README does have an example config

tzumainn:
@DanNiESh updated the code - let me know what you think!

esiclient/v1/orchestrator/openshift.py
```diff
@@ -0,0 +1,368 @@
...
+        raise OSAIException(
+            'Unknown method used when calling Assisted Installer')
+    if response.status_code not in [200, 201, 202, 204]:
+        raise OSAIException(
+            "Unexpected response from Assisted Installer (%s): %s"
+            % (response.status_code, response.reason))
+    return response
+
+
+def wait_for_nodes(infra_env_id, headers, num_nodes, target_status):
```
DanNiESh, line 47: Do you think adding a timeout might be useful? Sometimes nodes may have unexpected issues during booting?

esiclient/v1/orchestrator/openshift.py
```diff
@@ -0,0 +1,368 @@
...
+                response = call_assisted_installer_api(
+                    "infra-envs/%s/downloads/image-url" % infra_env_id,
+                    'get', headers)
+                image_url = response.json().get('url')
+                provisioning_network = neutron_client.find_network(
+                    provisioning_network_name)
+                print("provisioning nodes")
+                for node_name in nodes:
+                    node = ironic_client.node.get(node_name)
+                    if node.provision_state == 'available':
```
DanNiESh, line 232: Do we want to make sure that all nodes are in 'available' state before registering nodes?

esiclient/v1/orchestrator/openshift.py
```diff
@@ -0,0 +1,368 @@
...
+                    response = call_assisted_installer_api(
+                        "clusters/%s/" % cluster_id,
+                        'get', headers)
+                    status = response.json().get('status')
+                    if status == 'installing':
+                        status = status + " " + str(
+                            response.json().get('progress'))
+                    print("* installation status: %s" % status)
+                    if status == 'installed':
+                        waiting = False
```
DanNiESh, line 321: Same. Is a timeout needed?

DanNiESh (approved):
LGTM

PR #58, opened 2023-11-17
Initial iteration of ESI cluster orchestrator

tzumainn:
This command allows a user to provision nodes with a glance image or boot from a url, and configure networking.

### DIFF
--- a/esiclient/v1/orchestrator/cluster.py
+++ b/esiclient/v1/orchestrator/cluster.py
@@ -1,0 +1,224 @@
+#   Licensed under the Apache License, Version 2.0 (the "License"); you may
+#   not use this file except in compliance with the License. You may obtain
+#   a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#   WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#   License for the specific language governing permissions and limitations
+#   under the License.
+
+import concurrent.futures
+import json
+import logging
+
+from osc_lib.command import command
+from osc_lib.i18n import _
+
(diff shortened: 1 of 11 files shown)


joachimweyl:
@tzumainn is this ready for review?

esiclient/v1/orchestrator/cluster.py
```diff
@@ -0,0 +1,224 @@
...
+        with concurrent.futures.ThreadPoolExecutor() as executor:
+            for node_config in node_configs:
+                provisioning_type = \
+                    node_config['provisioning']['provisioning_type']
+                if provisioning_type not in self.PROVISIONING_METHODS:
+                    raise ESIOrchestrationException(
+                        "Unknown provisioning method %s" % provisioning_type)
+                nodes = node_config['nodes']['ironic_nodes']
+                for node in nodes:
+                    future = executor.submit(
```
DanNiESh, line 194: I wonder do we need a try-expection block here? Also cleanup resources (delete the port or set the node to available if there's an error happens)

DanNiESh: Can you print more information during node provisioning? Such as which node is being deployed. It might be useful when the execution time is too long and user has no idea about what's the current stage.

tzumainn: Oh, there are messages printed within the `provision_node` function.

I'm a bit undecided upon the other questions. Right now, if multiple nodes are provisioned and only one node has an error, the others will still provision to completion. I'm not sure I actually want to delete resources in that case. I think I'm also against auto-cleaning because it takes away the ability of the user to investigate the issue.

In any case, exceptions for individual node issues end up printed in stdout, which I - think? - is fine. I catch ESI-specific exceptions, and let the others just bubble up.

DanNiESh: Yeah, I agree.

DanNiESh (approved):
LGTM

PR #60, opened 2023-11-28
esi-node-console-connect

skanthed:


joachimweyl:
@skanthed is this PR ready for review?

esiclient/tests/unit/v1/test_node_console.py
```diff
@@ -0,0 +1,70 @@
...
+        }
+        self.node2 = {
+            "uuid": "11111111-2222-3333-4444-bbbbbbbbbbbb",
+            "name": "node2",
+            "console_enabled": True,
+            "console_info": {"type": "socat", "url": "tcp://10.10.10.10:1234"}
+            }
+        self.node3 = None
+
+    def test_take_action(self):
```
tzumainn, line 39: It looks like there are three separate test cases in here; you'll want to split them up, similarly to https://github.com/CCI-MOC/python-esiclient/blob/master/esiclient/tests/unit/v1/test_node_network.py#L127-L235, so that it's easy to see what's being tested with each result.

esiclient/tests/unit/v1/test_node_console.py
```diff
@@ -0,0 +1,70 @@
...
+from esiclient.v1 import node_console
+
+
+class TestNodeConsoleConnect(base.TestCommand):
+
+    def setUp(self):
+        super(TestNodeConsoleConnect, self).setUp()
+        self.cmd = node_console.NodeConsoleConnect(self.app, None)
+
+        self.node1 = {
```
tzumainn, line 25: Minor suggestion, but I might call this something like `self.node_console_1` just to ensure that it does not represent an actual node.

esiclient/tests/unit/v1/test_node_console.py
```diff
@@ -0,0 +1,70 @@
...
+            "uuid": "11111111-2222-3333-4444-aaaaaaaaaaaa",
+            "name": "node1",
```
tzumainn, line 27: This represents the return value of `openstack baremetal node console show`, right? If so, I don't think it returns a uuid or name; just the console values below.

esiclient/tests/unit/v1/test_node_console.py
```diff
@@ -0,0 +1,70 @@
...
+            'ERROR: No console info for node1',
+            self.cmd.take_action, parsed_args)
+
+        self.app.client_manager.baremetal.node.get_console.\
+            return_value = self.node2
+
+        arglist = ['node2']
+        parsed_args = self.check_parser(self.cmd, arglist, verifylist)
+
+        # result = self.cmd.take_action(parsed_args)
```
tzumainn, line 59: I think this is the only node that `succeeds`; are you not running the test?

esiclient/v1/node_console.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def take_action(self, parsed_args):
+        self.log.debug("take_action(%s)", parsed_args)
+
+        node = parsed_args.node
+
+        ironic_client = self.app.client_manager.baremetal
+
+        try:
+            output = ironic_client.node.get_console(node)
```
tzumainn, line 44: Is the try/except block needed here? If the node doesn't exist, is the output simply `None`, or does it throw an Exception? If the first, you don't need the try/except block; if the latter, then you probably don't want the if statement beneath.

esiclient/v1/node_console.py
```diff
@@ -0,0 +1,60 @@
...
+            if output is None:
+                raise
+
+        except Exception:
```
tzumainn, line 48: If the Exception you're catching is a specific one, it's best to catch the exact one.

esiclient/v1/node_console.py
```diff
@@ -0,0 +1,60 @@
...
+            raise exceptions.CommandError("ERROR: Node unknown")
+
+        console_show_output = json.loads(json.dumps(output))
+
+        console_info = console_show_output['console_info']
+
+        if console_info is None:
+            raise exceptions.CommandError("ERROR: No console info for "+node)
```
tzumainn, line 56: I think it's slightly preferred to do something like `ERROR: No console info for %s" % node`.

It may also be useful for the error message to explain what the user should do: run `"openstack baremetal node console enable %s" % node`

tzumainn (changes requested):
Thanks for the PR! A few comments, but I think they'll be easy to fix.

esiclient/v1/node_console.py
skanthed: "output" does not catch any value because that function is simply throwing the error not raising the error which means rest of the lines will not be executed. if statement is just for the test case scenario because we assigned none as a return value for the unknown node test case.

tzumainn: Okay - we do want to catch the exact Error then; if you run the command with `--debug` you should see it.

We shouldn't have the if statement for the test case scenario. You can have the mock console command throw an error using `side_effect` - https://docs.python.org/3/library/unittest.mock.html

esiclient/v1/node_console.py
skanthed: okay made it more meaningful.

esiclient/v1/node_console.py
skanthed: Its a generic exception message! We are just catching the exception from whatever it's returning from console show.

tzumainn: It's not `ironicclient.common.apiclient.exceptions.NotFound`?

tzumainn: Actually, after thinking about it a bit more - I think it's fine to simply not have a try/catch block here. Then the original Ironic "node not found" exception bubbles up, and that's perfectly comprehensible to the end user.

esiclient/tests/unit/v1/test_node_console.py
skanthed: We should get some message after it's connected to TCP. But it's returning none always. Looking for some alternative approach to get correct message connection refused or connected.
If you have any suggestion @tzumainn  please let me know. putting out some commented code.

skanthed: okay did something. It's working!

esiclient/v1/node_console.py
skanthed: Used the specific exception!

esiclient/v1/node_console.py
skanthed: I know the exception already that's not the issue.

esiclient/v1/node_console.py
```diff
@@ -0,0 +1,65 @@
...
+        self.log.debug("take_action(%s)", parsed_args)
+
+        node = parsed_args.node
+
+        ironic_client = self.app.client_manager.baremetal
+
+        try:
+            output = ironic_client.node.get_console(node)
+            if output is None:
+                raise NotFound
```
tzumainn, line 47: I think this comment may have been lost earlier, but `output` is never actually None, correct? This is only for test purposes? If so, then this `if` statement should be removed, and the test reworked.

esiclient/tests/unit/v1/test_node_console.py
```diff
@@ -0,0 +1,102 @@
...
+        parsed_args = self.check_parser(self.cmd, arglist, verifylist)
+
+        result = self.cmd.take_action(parsed_args)
+
+        self.assertEqual(0, result)
+        mock_system.assert_called_once
+
+    @mock.patch('esiclient.v1.node_console.os.system',
+                return_value=256, autospec=True)
+    def test_take_action_invalid_tcp_url(self, mock_system):
```
tzumainn, line 58: You don't really need this test; it's pretty much the same as the test before it, except you're changing the mocked return value of `os.system`, which is passed directly to the user.

esiclient/tests/unit/v1/test_node_console.py
```diff
@@ -0,0 +1,102 @@
...
+            exceptions.CommandError,
+            'ERROR: No console info for node_console_3. '
+            'Run openstack baremetal node console '
+            'enable for given node',
+            self.cmd.take_action, parsed_args)
+
+    def test_take_action_unknown_console_node(self):
+
+        self.app.client_manager.baremetal.node.get_console.\
+            return_value = None
```
tzumainn, line 92: Does `get_console` ever actually return `None`? I thought it'd only return a dict or an exception. If so, you can remove this test case and get rid of the if statement I mention below.

esiclient/v1/node_console.py
```diff
@@ -0,0 +1,65 @@
...
+
+        except NotFound:
+            raise exceptions.CommandError("ERROR: Node not found")
```
tzumainn, line 50: I don't think you need to catch the `NotFound` exception; you're essentially catching one error just to throw the new one. I think it's fine for the original `NotFound` exception to be thrown to the user.

esiclient/tests/unit/v1/test_node_console.py
skanthed: okay

esiclient/tests/unit/v1/test_node_console.py
skanthed: No. I am removing it.

tzumainn (approved):
Looks good - thanks!

PR #61, opened 2023-12-07
Add undeploy commands

tzumainn:
These undeploy commands allow users to cleanup orchestrated clusters.


esiclient/v1/orchestrator/cluster.py
```diff
@@ -222,3 +223,80 @@ def take_action(self, parsed_args):
 
         return ["Node", "Port", "Network", "Fixed IP",
                 "Floating Network", "Floating IP"], data
+
+
+class Undeploy(command.Command):
+    """Undeploy a cluster from ESI nodes"""
+
+    log = logging.getLogger(__name__ + ".Undeploy")
+    REQUIRED_FIELDS = ['nodes', 'private_network_name', 'api_vip',
```
DanNiESh, line 232: I think REQUIRED_FIELDS is not used to check fields in config file in this class.

esiclient/v1/orchestrator/cluster.py
```diff
@@ -222,3 +223,80 @@ def take_action(self, parsed_args):
...
+                    if port:
+                        print("   * %s" % port_name)
+                        neutron_client.delete_port(port.id)
+
+        non_uuid_node_configs = [node_config for node_config in node_configs
+                                 if 'node_uuids' not in node_config['nodes']]
+        if len(non_uuid_node_configs) > 0:
+            print("* %s node configs were skipped, as they do not"
+                  " specify specific nodes" % len(non_uuid_node_configs))
+            print("   * these nodes and ports will have to be removed"
```
DanNiESh, line 293: How can user keep track of these unspecified nodes? Should we provide detailed instructions here?

esiclient/v1/orchestrator/cluster.py
```diff
@@ -222,3 +223,80 @@ def take_action(self, parsed_args):
...
+            print("* undeploying %s of %s node configs" % (
+                count, len(uuid_node_configs)))
+            node_uuids = node_config['nodes']['node_uuids']
+            network_config = node_config['network']
+            network_uuid = network_config.get('network_uuid', None)
+            network = neutron_client.find_network(network_uuid)
+            for node in node_uuids:
+                print("   * %s" % node)
+                ironic_client.node.set_provision_state(node, 'deleted')
+                print("   * waiting for nodes to start undeploy before"
```
DanNiESh, line 272: Will you get an error if deleting a port while a node is still under deleting?

tzumainn: Nope - but it takes ~10 seconds for Ironic to detach the port itself, so I figure there's no harm in waiting!

tzumainn:
@DanNiESh thanks for your comments! Answers in-line. Longer-term, I'm hoping to implement some changes to add some cluster metadata so that removing resources should be much easier.

PR #64, opened 2024-05-10
Update `node network list` to include floating IP port forwarding

tzumainn:
Note that there is no way to grab all port forwarding information across all floating IPs; as a result, it has to be done one API command at a time.


tzumainn:
@DanNiESh could you take a look at this when you have a chance? Thanks!

DanNiESh (approved):

PR #65, opened 2024-06-26
node network list now calls from esisdk

ajamias:
To centralize logic, the `esi node network list` will call from esisdk.

### DIFF
--- a/esiclient/tests/unit/v1/test_node_network.py
+++ b/esiclient/tests/unit/v1/test_node_network.py
@@ -24,7 +24,6 @@ class TestList(base.TestCommand):
 
     def setUp(self):
         super(TestList, self).setUp()
-        self.cmd = node_network.List(self.app, None)
 
         self.port1 = utils.create_mock_object({
             "uuid": "port_uuid_1",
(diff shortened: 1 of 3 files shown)


esiclient/tests/unit/v1/test_node_network.py
```diff
@@ -26,229 +26,14 @@ def setUp(self):
...
-        )
-        self.assertEqual(expected, results)
-        self.app.client_manager.baremetal.port.list.\
-            assert_called_once_with(detail=True)
-
-    def test_take_action_neutron_port_does_not_exist(self):
-        self.app.client_manager.baremetal.port.list.\
-            return_value = [self.port4]
-
+    def test_get_parser(self):
```
tzumainn, line 29: There's no real need to test this; however, we do need to test `take_action`. Since it's a unit test, what happens is you mock any outside function calls that you make (previously the various openstack `self.app.client_manager.XXXX.YYYYY`) and hard code the results. Then you verify the results, and that the appropriate function calls are made. You also have to test each branch in a conditional, which is why there were so many tests here; however since the updated code shifts all that into into the `esisdk`, we can eliminate those test cases.

That means in this case, we just have to mock `node.network_list`. As a matter of thoroughness though, I'd mock the return value to cover a variety of cases - trunk ports, floating IPs, floating IPs with port forwarding, as well as baremetal ports with nothing associated with them.

If you haven't done so yet, I think you should be able to create trunk ports and floating IPs (the ESI documentation should explain how). I'd recommend doing so and running `esi network list` with the current packaged version of `python-esiclient` to see what it looks like.

tzumainn (changes requested):
I can't fully test this, since there's an error I'm tracing back to `esisdk` (which I'll comment on there). But broadly, I think it looks good; however you'll want to update the unit test.

esiclient/tests/unit/v1/test_node_network.py
```diff
@@ -24,7 +24,7 @@ class TestList(base.TestCommand):
 
     def setUp(self):
         super(TestList, self).setUp()
-        self.cmd = node_network.List(self.app, None)
+        self.maxDiff = None
```
tzumainn, line 27: I don't think this is needed (at least from my testing).

tzumainn (changes requested):
One minor comment; looks good to me otherwise!

esiclient/tests/unit/v1/test_node_network.py
ajamias: Ah that was for my debugging, I'll get rid of it now

tzumainn (approved):

tzumainn:
Thanks!

PR #66, opened 2024-07-18
set env variable for api version in python context

sheldor1510:
- fix for https://github.com/CCI-MOC/esi/issues/555

### DIFF
--- a/esiclient/plugin.py
+++ b/esiclient/plugin.py
@@ -12,6 +12,7 @@
 #
 
 import logging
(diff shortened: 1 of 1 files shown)


tzumainn (changes requested):
This works well! The only request I have is to make the baremetal API version a constant at the top of the file; that will make it more apparent and easier to change in the future if needed.

tzumainn (approved):

PR #70, opened 2024-12-16
Fix microversion compatibility issue for OpenStack Baremetal API

skanthed:
Closes cci-moc/esi#669

### DIFF
--- a/esiclient/tests/unit/v1/mdc/test_mdc_node_baremetal.py
+++ b/esiclient/tests/unit/v1/mdc/test_mdc_node_baremetal.py
@@ -11,89 +11,55 @@
 #   under the License.
 #
 
-import mock
-import munch
+from unittest import mock
+import openstack.baremetal.v1.node
+import openstack.connection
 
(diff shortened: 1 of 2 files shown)


skanthed:
@tzumainn I’m unable to select you as reviewer to this PR directly. Could you please check if I can be given the necessary permissions to do so for this repo?

tzumainn:
Ah,  I don't think this is the change we need - first, I think our version of Ironic doesn't go up to 1.72; second, the core issue is that the version set there isn't being passed into the call to Ironic; I suspect something needs to be set in the esisdk. Make sure you test in a clean python environment, first verifying that if you unset `OS_BAREMETAL_API_VERSION` then a command like `openstack esi node network list` will fail; then verifying it passes with whatever changes you make!

larsks (changes requested):
I agree with Mainn's comment; just marking this reviewed so I don't get reminders for it.

joachimweyl:
@skanthed what are the next steps?

PR #72, opened 2024-12-20
Teach node network detach about --all option

larsks:
Update python-esiclient for compatability with the changes introduced in
cci-moc/esisdk#15. This adds support for the `--all` option (delete all
ports), and also permits the `--port` option to be used multiple times
to delete multiple individual ports.

### DIFF
--- a/esiclient/tests/unit/v1/test_node_network.py
+++ b/esiclient/tests/unit/v1/test_node_network.py
@@ -1018,16 +1018,6 @@ class TestDetach(base.TestCommand):
         self.node = utils.create_mock_object(
             {"id": "node_uuid_1", "name": "node1", "provision_state": "active"}
         )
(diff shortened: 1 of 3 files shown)


larsks:
Sorry for the noise; I prematurely tagged some reviewers on this. It still needs tests so is not actually ready to merge.

larsks:
Note that this requires https://github.com/CCI-MOC/esisdk/pull/15

tzumainn (commented):
This looks good as well. I can create an esisdk release so you can update the requirements.txt here, unless you think there's a better option?

larsks:
I think cutting a new esisdk release is probably the best course of action.

tzumainn:
Okay - esisdk 1.4 should have the code you need; update `requirements.txt` and I can merge this!

tzumainn (approved):
Looks good - thanks!

PR #73, opened 2025-01-10
Add `--long` option to `openstack esi node network list`

tzumainn:
This option adds resource UUIDs to the output.


DanNiESh (approved):
LGTM

PR #74, opened 2025-01-14
Remove opportunities for skew from PR 73

larsks:
This was going to be a review of #73, but I was too slow.

The structure introduced to support the `--long` option means we need to
maintain the list of field names and corresponding data in two different
places, which opens up opportunities for skew (e.g, someone adds a new
field to the output without updating the behavior for `--long`).

This refactors the handling of `--long` so that it only ever augments the
data returned.


larsks:
Looks like I need to update some tests.

larsks:
There, all better.

tzumainn (approved):
yeah, you're right - I was trying to keep a "logical" ordering of columns with node and port attributes together, but I think this issue is exactly why openstack `--long` commands don't order their columns

PR #84, opened 2025-01-24
Make "esi mdc baremetal node list" work

larsks:
OpenStackConfig.get_all() will fail with a keystone exception if there are
any OS_* variables set in the environment [1]. This is easy to trigger if,
for example, you have set OS_BAREMETAL_API_VERSION in your environment to
work around [2].

We can avoid the problem by removing the call to get_all(), and
instead initializing the list of clouds using get_cloud_names().

This commit makes a number of additional changes to
esiclient/v1/mdc/mdc_node_baremetal.py:

- Several attribute names appear to have changed since this code was
  written (node.uuid is now node.id; node.instance_uuid is now
  node.instance_id).

- Replace single-letter variable names with more meaningful names.

- Add the '--ignore-invalid' command, which allows the command to continue
  in the event that it cannot successfully connect to one or more clouds.

[1]: https://bugs.launchpad.net/openstacksdk/+bug/2096621
[2]: https://github.com/CCI-MOC/esi/issues/669

### DIFF
--- a/esiclient/tests/unit/v1/mdc/test_mdc_node_baremetal.py
+++ b/esiclient/tests/unit/v1/mdc/test_mdc_node_baremetal.py
@@ -11,89 +11,55 @@
 #   under the License.
 #
 
-import mock
-import munch
+from unittest import mock
+import openstack.baremetal.v1.node
+import openstack.connection
 
 from esiclient.tests.unit import base
 from esiclient.v1.mdc import mdc_node_baremetal
 
 
+@mock.patch("esiclient.v1.mdc.mdc_node_baremetal.openstack.connect")
+@mock.patch("openstack.config.loader.OpenStackConfig.get_cloud_names")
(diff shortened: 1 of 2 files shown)


larsks:
This PR was sort of an accident, because I was looking at setup.cfg and found myself asking, "what the heck is this `mdc` subcommand?".

joachimweyl:
@tzumainn saw this while I was reviewing PRs.

tzumainn:
@larsks Oh, ha, it was a sort of experimental thing created in the wilder days of ESI that has never become useful. I like your changes; however I also think maybe this code should just be removed as irrelevant - what do you think?

larsks:
I guess I would say (a) merge this code then (b) submit a new pr removing the command. This way the fixed code exists if somebody ever wants to go looking for it.

tzumainn:
sounds good!

tzumainn (approved):

PR #86, opened 2025-03-31
Fix the baremetal microversion issue

skanthed:
Sets the default `OS_BAREMETAL_API_VERSION` earlier in plugin.py to ensure it's applied before OpenStackClient initializes.
Fixes microversion mismatch by preventing fallback to unsupported default (1.95) during CLI execution.

### DIFF
--- a/esiclient/plugin.py
+++ b/esiclient/plugin.py
@@ -21,6 +21,9 @@ LOG = logging.getLogger(__name__)
 DEFAULT_BAREMETAL_API_VERSION = "1.69"
 DEFAULT_ESICLIENT_API_VERSION = "1"
 
+if "OS_BAREMETAL_API_VERSION" not in os.environ:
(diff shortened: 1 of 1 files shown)


larsks (changes requested):
Please add a commit message and pull request description. What issue does this address? How does it implement the correction?

tzumainn (approved):
Looks reasonable to me - I tried moving the if statement into various places, and can confirm that none of it works elsewhere. There is almost definitely a more elegant solution somewhere, but I think that may require digging deep into the openstackclient code.

tzumainn:
@skanthed you may want to squash the commits to get rid of the merge commit, though

skanthed:
@tzumainn Squashed it!

tzumainn:
Ah, I think you may have to rebase from master first! It looks like the ubuntu OS update is manually applied here.

larsks:
And please copy the PR description to the commit message, which is still empty.

larsks (approved):

tzumainn (approved):
Looks good!

PR #87, opened 2025-04-18
Update test OS to Ubuntu 24.04

tzumainn:
Ubuntu 20.04 was deprecated on 2025-04-15; this updates the version of Ubuntu we use.


skanthed (approved):
LGTM

PR #88, opened 2025-05-13
Improve error handling in ..esi trunk create

skanthed:
Allows Neutron's  error messages to surface properly during invalid network references.
- Use `ignore_missing=False` in find_network to raise OpenStack SDK errors
- Removes silent fallback to None

### DIFF
--- a/esiclient/tests/unit/v1/test_trunk.py
+++ b/esiclient/tests/unit/v1/test_trunk.py
@@ -253,7 +253,7 @@ class TestCreate(base.TestCommand):
         self.assertEqual(expected, results)
         self.app.client_manager.network.find_network.assert_has_calls(
             [
-                mock.call("network1"),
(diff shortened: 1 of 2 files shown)


larsks:
The associated issue for this PR is https://github.com/CCI-MOC/esi/issues/664, which provides two examples in which we are hiding relevant error messages from the user. It looks like this PR only addresses the second of those two situations (the failure of `esi trunk create` due to a missing network). 

Were you able to reproduce the first error?

skanthed:
Yes I was and will be addressed in the new PR.

skanthed:
node network attach fix - [https://github.com/CCI-MOC/esisdk/pull/25](https://github.com/CCI-MOC/esisdk/pull/25)

tzumainn (approved):
Looks good - thanks!

